PR #7, opened 2022-01-13
INC-49: Display active case load and link to change

xoen:
Similarly to how it's done in other DPS services.


Mjwillis (approved):
LGTM - but API needs changing to a new one!

PR #8, opened 2022-01-20
INC-50: Ability to select current location (aka "wing")

xoen:
- User is asked to chose a location - if they haven't already - when looking at the behaviour entries page
- Active location is checked against current case note/prison to be sure when a user change case note they're asked to chose a location from the newly selected prison


integration_tests/mockApis/prisonApi.ts
```diff
@@ -0,0 +1,76 @@
...
+          },
+        ],
+      },
+    })
+  },
+  stubGetAgencyLocations: (): SuperAgentRequest => {
+    return stubFor({
+      request: {
+        method: 'GET',
+        urlPattern: '/prisonApi/api/agencies/MDI/locations',
```
Mjwillis, line 44: Wrong Endpoint:  Should be `{{prison_api}}/api/users/me/locations`

server/data/prisonApi.ts
```diff
@@ -8,14 +8,60 @@ interface CaseLoad {
...
-    return this.get({ path: '/users/me/caseLoads' }) as Promise<Array<CaseLoad>>
+    return this.get({ path: '/api/users/me/caseLoads' }) as Promise<Array<CaseLoad>>
+  }
+
+  async getAgencyLocations(agencyId: string): Promise<Array<Location>> {
+    const locations = (await this.get({ path: `/api/agencies/${agencyId}/locations` })) as Array<Location>
+
+    // Only return occupied wings
+    return locations.filter(location => {
+      return location.locationType === 'WING' && location.currentOccupancy > 0
```
Mjwillis, line 62: Not sure if we should add a new endpoint to just get wings or you can simply call `{{prison_api}}/api/users/me/locations` which will give you the current users wing locations for the prison they have selected

server/data/prisonApi.ts
```diff
@@ -8,14 +8,60 @@ interface CaseLoad {
   type: string
 }
 
+// Possibly not an exhaustive list
```
Mjwillis, line 11: Don't think this is necessary, only interested in WING

server/data/prisonApi.ts
```diff
@@ -8,14 +8,60 @@ interface CaseLoad {
...
+  | 'RTU'
+  | 'VIDE'
+  | 'TRAI'
+  | 'LAND'
+  | 'SPUR'
+  | 'HOLD'
+
+interface Location {
+  locationId: number
+  locationType: LocationType
```
Mjwillis, line 39: Make this is string

server/data/prisonApi.ts
```diff
@@ -8,14 +8,60 @@ interface CaseLoad {
...
+}
+
 class PrisonApi extends RestClient {
   constructor(token: string) {
     super('HMPPS Prison API', config.apis.hmppsPrisonApi, token)
   }
 
   async getUserCaseLoads(): Promise<Array<CaseLoad>> {
-    return this.get({ path: '/users/me/caseLoads' }) as Promise<Array<CaseLoad>>
+    return this.get({ path: '/api/users/me/caseLoads' }) as Promise<Array<CaseLoad>>
```
Mjwillis, line 54: This one should be moved (in a different story) https://nomis-user-dev.aks-dev-1.studio-hosting.service.justice.gov.uk/swagger-ui/index.html#/
`/me/caseloads`. gives:
```{
  "username": "TESTUSER1",
  "active": true,
  "accountType": "GENERAL",
  "activeCaseload": {
    "id": "WWI",
    "name": "WANDSWORTH (HMP)"
  },
  "caseloads": [
    {
      "id": "WWI",
      "name": "WANDSWORTH (HMP)"
    }
  ]
}```

server/data/prisonApi.ts
Mjwillis: Not sure about the occupancy check - is that a requirement?

Mjwillis (approved):
Approved but we'll want to change some of the APIs

server/data/prisonApi.ts
xoen: Yeah, I was tempted to change the `locationType` field type to string as well. I'll change as it's simpler👌

integration_tests/mockApis/prisonApi.ts
xoen: wouldn't this other `/api/users/me/locations` endpoint return all locations accessible to the user tho? Including ones for a case load which is not the active one?

server/data/prisonApi.ts
xoen: The occupancy check wasn't a requirement but when I was playing with the API we saw that some wings were repeated/similar. So this is to avoid showing empty wings. At this stage not much point in allowing a user to select a wing which has no data IMHO.

xoen:
Thanks for the review Michael. As discussed offline, I'm merging this as it is and I'll tweak/experiment with these comments separately ☺️

server/data/prisonApi.ts
xoen: Yep, I added ticket [INC-86](https://dsdmoj.atlassian.net/browse/INC-86) for this ☺️

PR #10, opened 2022-01-28
Increased Incentives API timeouts to 60s

xoen:
Some requests, expecially while API cache is not warm, can take a bit long
so we're increasing the timeouts to try and reduce/avoid them.


Mjwillis (dismissed):
LGTM

Mjwillis (approved):
LGTM

PR #17, opened 2022-01-31
INC-160 Add links to DPS

Mjwillis:


integration_tests/mockApis/incentivesApi.ts
```diff
@@ -21,7 +21,8 @@ export default {
     return stubFor({
       request: {
         method: 'GET',
-        urlPattern: '/incentivesApi/incentives-summary/prison/MDI/location/MDI-42',
+        urlPattern:
+          '/incentivesApi/incentives-summary/prison/MDI/location/MDI-42\\?sortBy=DAYS_SINCE_LAST_REVIEW&sortDirection=DESC',
```
Mjwillis, line 25: Just added this for now to indicate sorting

server/data/incentivesApi.test.ts
```diff
@@ -26,7 +26,9 @@ describe('IncentiveApi', () => {
       const locationId = 'ABC-1'
       const apiResponse = getTestIncentivesLocationSummary({ prisonId, locationId })
       incentivesApi
-        .get(`/incentives-summary/prison/${prisonId}/location/${locationId}`)
+        .get(
+          `/incentives-summary/prison/${prisonId}/location/${locationId}?sortBy=DAYS_SINCE_LAST_REVIEW&sortDirection=DESC`
```
Mjwillis, line 30: Sample sort example, will replace with params

xoen: I left the sorting out as the default seems to be sorting by name (both in terms of Incentives API and in terms of requirements apparently) - but adding this here would mean the default ordering would now be wrong.

server/views/pages/incentives-table.njk
```diff
@@ -69,13 +69,13 @@
...
             { html: '<a href="' + dpsHome + '/prisoner/' + prisoner.prisonerNumber + '" class="govuk-link">' + prisoner.lastName + ', ' + prisoner.firstName + ' (' + prisoner.prisonerNumber + ')</a>' },
-            { text: prisoner.daysOnLevel },
-            { text: prisoner.daysSinceLastReview },
-            { text: prisoner.positiveBehaviours },
-            { text: prisoner.incentiveEncouragements },
-            { text: prisoner.negativeBehaviours },
-            { text: prisoner.incentiveWarnings },
-            { text: prisoner.provenAdjudications }
+            { html: '<a href="' + dpsHome + '/prisoner/' + prisoner.prisonerNumber + '/incentive-level-details?agencyId=' + entries.prisonId + '&incentiveLevel=' + level.levelDescription + '" class="govuk-link">' + prisoner.daysOnLevel + '</a>' },
+            { html: '<a href="' + dpsHome + '/prisoner/' + prisoner.prisonerNumber + '/incentive-level-details?agencyId=' + entries.prisonId + '&incentiveLevel=' + level.levelDescription + '" class="govuk-link">' + prisoner.daysSinceLastReview + '</a>' },
```
xoen, line 73: This link and the one above are the same (AFAICT) - maybe it's worth having it in a variable - or maybe it's overkill.

xoen (changes requested):
This looks good but the API call to Incentives API changed the default order from "by name" (which I think it's what they want by default) to "by days since last review DESC" - I think this should be changed back.

xoen (approved):

PR #19, opened 2022-02-01
INC-167: Tweaks to behaviour entries table's links

xoen:
- Prisoner name links to DPS 'case notes' tab (and with `fromDate` set to 3 months ago)
- links to DPS Incentive Level Details don't filter by level
- Adjudications links don't have the `fromDate` filter (as this is likely to not show proven adjudications because the date they were recorded is older than 3 months)

Also, showing Prisoner Number below their last name/first name as in prototype(s).


Mjwillis (approved):
Lgtm

PR #22, opened 2022-02-04
INC-163 Timeout fix

Mjwillis:
- Correctly set timeout for API calls
- Add number to table tabs
- Show all prisoner locations

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -15,11 +15,11 @@ function get<T>(name: string, fallback: T, options = { requireInProduction: fals
 const requiredInProduction = { requireInProduction: true }
 
 export class AgentConfig {
-  maxSockets: 100
+  timeout: number
 
-  maxFreeSockets: 10
-
-  freeSocketTimeout: 30000
+  constructor(timeout = 8000) {
+    this.timeout = timeout
+  }
 }
 
 export interface ApiConfig {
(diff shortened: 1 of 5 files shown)


server/config.ts
```diff
@@ -15,11 +15,11 @@ function get<T>(name: string, fallback: T, options = { requireInProduction: fals
 const requiredInProduction = { requireInProduction: true }
 
 export class AgentConfig {
-  maxSockets: 100
+  timeout: number
 
-  maxFreeSockets: 10
```
Mjwillis, line 20: These values are ignored

server/data/prisonApi.ts
```diff
@@ -29,15 +29,15 @@ class PrisonApi extends RestClient {
...
-    return this.get({ path: '/api/users/me/caseLoads' }) as Promise<Array<CaseLoad>>
+    return (await this.get({ path: '/api/users/me/caseLoads' })) as Promise<Array<CaseLoad>>
   }
 
   async getUserLocations(): Promise<Array<Location>> {
     const locations = (await this.get({ path: `/api/users/me/locations` })) as Array<Location>
 
     // Only return occupied wings
     return locations.filter(location => {
-      return location.locationType === 'WING' && location.currentOccupancy > 0
```
Mjwillis, line 40: Should show any location with prisoners in

server/views/pages/incentives-table.njk
```diff
@@ -98,7 +98,7 @@
     {% for level in entries.incentiveLevelSummary %}
         {% set _ = tabsItems.push(
             {
-                label: level.levelDescription,
+                label: level.levelDescription + ' (' + level.numberAtThisLevel + ')',
```
Mjwillis, line 101: UI Improvement

petergphillips (approved):
LGTM

server/data/prisonApi.ts
```diff
@@ -29,15 +29,15 @@ class PrisonApi extends RestClient {
   }
 
   async getUserCaseLoads(): Promise<Array<CaseLoad>> {
-    return this.get({ path: '/api/users/me/caseLoads' }) as Promise<Array<CaseLoad>>
+    return (await this.get({ path: '/api/users/me/caseLoads' })) as Promise<Array<CaseLoad>>
```
ushkarev, line 32: you don't actually need to do this. because the function is async so is awaited elsewhere, adding await just before a return means there's a double-await on the same promise. this explains it better than i do: https://eslint.org/docs/rules/no-return-await

additionally, adding the await keyword means that the type assertion at the end ought to be `Array<CaseLoad>` because the promise has already been resolved.

PR #23, opened 2022-02-08
INC-86 Use different api for user caseload.

Mjwillis:
Signed-off-by: mjwillis <michael.willis@digital.justice.gov.uk>

### DIFF
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -40,9 +40,9 @@ class IncentivesApi extends RestClient {
   }
 
   async getLocationSummary(agencyId: string, locationPrefix: string): Promise<IncentivesLocationSummary> {
-    return this.get({
+    return (await this.get({
       path: `/incentives-summary/prison/${agencyId}/location/${locationPrefix}?sortBy=NAME&sortDirection=ASC`,
-    }) as Promise<IncentivesLocationSummary>
+    })) as Promise<IncentivesLocationSummary>
   }
 }
 
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -39,10 +39,10 @@ class IncentivesApi extends RestClient {
     super('HMPPS Incentives API', config.apis.hmppsIncentivesApi, systemToken)
   }
 
-  async getLocationSummary(agencyId: string, locationPrefix: string): Promise<IncentivesLocationSummary> {
-    return (await this.get({
+  getLocationSummary(agencyId: string, locationPrefix: string): Promise<IncentivesLocationSummary> {
+    return this.get({
       path: `/incentives-summary/prison/${agencyId}/location/${locationPrefix}?sortBy=NAME&sortDirection=ASC`,
-    })) as Promise<IncentivesLocationSummary>
+    }) as Promise<IncentivesLocationSummary>
   }
 }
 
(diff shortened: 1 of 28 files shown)


server/data/incentivesApi.ts
```diff
@@ -40,9 +40,9 @@ class IncentivesApi extends RestClient {
   }
 
   async getLocationSummary(agencyId: string, locationPrefix: string): Promise<IncentivesLocationSummary> {
-    return this.get({
+    return (await this.get({
```
ushkarev, line 43: you don't actually need to do this. because the function is async so is awaited elsewhere, adding await just before a return means there's a double-await on the same promise. this explains it better than i do: https://eslint.org/docs/rules/no-return-await

alternatively, adding the new await keyword means that the type assertion at the end ought to simply be `IncentivesLocationSummary` because the promise has already been resolved

server/data/nomisUserRolesApi.ts
```diff
@@ -0,0 +1,35 @@
...
+    super('NOMIS User Roles API', config.apis.nomisUserRolesApi, token)
+  }
+
+  async getUserCaseloads(): Promise<UserCaseload> {
+    const userCaseload = (await this.get({ path: '/me/caseloads' })) as UserCaseload
+
+    const caseloads =
+      userCaseload.caseloads != null
+        ? userCaseload.caseloads.filter(caseload => {
+            return caseload.id !== 'NWEB'
```
xoen, line 25: What's special about this `NWEB` prison?

server/routes/imageRouter.ts
```diff
@@ -0,0 +1,22 @@
+import { Router } from 'express'
+import csurf from 'csurf'
+
+const testMode = process.env.NODE_ENV === 'test'
+
+export default function imageRouter(): Router {
```
xoen, line 6: Shouldn't this router still use the authentication router? I'm assuming only authenticated users can get images right?

(sorry if this is done somewhere else)

server/data/nomisUserRolesApi.ts
```diff
@@ -0,0 +1,35 @@
+import config from '../config'
+import RestClient from './restClient'
+
+interface Caseload {
+  id: string
+  name: string
+}
+
+interface UserCaseload {
```
xoen, line 9: This is a minor thing but maybe the name should be plural, e.g. `UserCaseloads`, so that it's more explicit this is returning multiple caseloads.

server/routes/imageRouter.ts
Mjwillis: Yes only auth user can get images, this still applies

server/data/incentivesApi.ts
Mjwillis: @ushkarev I've removed the async from this and bubbled up the Promises

server/data/nomisUserRolesApi.ts
Mjwillis: NWEB is a special caseload which is being returned. But I've remove it from the results in this PR https://github.com/ministryofjustice/nomis-user-roles-api/pull/94

server/data/nomisUserRolesApi.ts
Mjwillis: Ok will do a future commit @xoen

PR #25, opened 2022-02-11
INC-162: unnecessary API calls when getting prisoner images

xoen:
The prisoner images route was initially using the `standardRouter` which
was populating the current user/caseloads making additional requests.

This is useful when showing a page but a waste when a browser is getting
prisoner images' raw data.

That route now uses a simpler `imageRouter` which doesn't use the
`populateUser` middleware...however, these API calls were still happening.

The reason for this is order of the routes wiring. The `/` route (which
still uses the `standardRouter`) was still running because it was more
generic. Move it last means these unnecessary requests are not made anymore.

NOTE: I've added a regression test - which would fail if the
`app.use('/', ...` is moved back above the prisoner images route) -
and also a test for the authentication requirement of this endpoint.

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -38,10 +38,10 @@ export default function createApp(userService: UserService): express.Application
   app.use(authorisationMiddleware())
 
   // App routes
-  app.use('/', homeRoutes(standardRouter(userService)))
   app.use('/select-location', selectLocationRoutes(standardRouter(userService)))
   app.use('/incentive-summary/:locationPrefix', incentivesTableRoutes(standardRouter(userService)))
   app.use('/prisoner-images/:imageId.jpeg', prisonerImagesRoutes(imageRouter()))
+  app.use('/', homeRoutes(standardRouter(userService)))
 
   app.use((req, res, next) => next(createError(404, 'Not found')))
   app.use(errorHandler(process.env.NODE_ENV === 'production'))
--- a/server/routes/prisonerImages.test.ts
+++ b/server/routes/prisonerImages.test.ts
@@ -3,17 +3,22 @@ import request from 'supertest'
 
 import { appWithAllRoutes } from './testutils/appSetup'
 import { PrisonApi } from '../data/prisonApi'
+import UserService from '../services/userService'
 
 jest.mock('../data/prisonApi')
+jest.mock('../services/userService')
 
 let app: Express
 let prisonApi: jest.Mocked<PrisonApi>
+let userService: jest.Mocked<UserService>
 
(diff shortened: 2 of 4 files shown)


server/app.ts
```diff
@@ -38,10 +38,10 @@ export default function createApp(userService: UserService): express.Application
   app.use(authorisationMiddleware())
 
   // App routes
-  app.use('/', homeRoutes(standardRouter(userService)))
   app.use('/select-location', selectLocationRoutes(standardRouter(userService)))
   app.use('/incentive-summary/:locationPrefix', incentivesTableRoutes(standardRouter(userService)))
   app.use('/prisoner-images/:imageId.jpeg', prisonerImagesRoutes(imageRouter()))
+  app.use('/', homeRoutes(standardRouter(userService)))
```
xoen, line 44: Move this after the `/prisoner-images/:imageId.jpeg` route avoid the unnecessary calls.

server/routes/prisonerImages.test.ts
```diff
@@ -36,4 +41,13 @@ describe('GET /prisoner-images/:imageId.jpeg', () => {
         expect(res.text).toEqual('image 123 data')
       })
   })
+
+  it('does not trigger getUser()', () => {
+    return request(app)
+      .get(`/prisoner-images/123.jpeg`)
+      .expect('Content-Type', /images\/jpeg/)
+      .expect(res => {
+        expect(userService.getUser).toBeCalledTimes(0)
```
xoen, line 50: Unit test to check that these calls don't happen when getting an image.

(will fail if moving the `/` route above the images route [here](https://github.com/ministryofjustice/hmpps-incentives-ui/pull/25/files#diff-47e00f44a4610e3f7bcaad8433c7bccf2ae050a627a3399de9553d7c910cd7b0R94))

ushkarev: this sounds fair enough, but given that the routing for tests is setup in a different place from the normal app's routing, it's still possible to accidentally change the real router and have this test pass

ushkarev (approved):

server/routes/prisonerImages.test.ts
xoen: You're right. I think at the moment this is a risk (not specific of this test)

I don't have a solution for this yet, but I think we may want to experiment with ways to improve separately.

Maybe we could experiment with the way routing is wired together at some point (e.g. have only one router mounted into `/` so that the `appSetup` module doesn't need to be updated. But when I played with this idea briefly everything broke 😬

PR #27, opened 2022-02-14
Add Global protect IP

Mjwillis:


ushkarev (approved):

helm_deploy/hmpps-incentives-ui/values.yaml
```diff
@@ -51,6 +51,7 @@ generic-service:
       REDIS_AUTH_TOKEN: "auth_token"
 
   allowlist:
+    global-protect: "35.176.93.186/32"
     office: "217.33.148.210/32"
```
ushkarev, line 55: perhaps we should also add 102PF?
https://github.com/ministryofjustice/moj-ip-addresses/blob/main/moj-cidr-addresses.yml#L67

PR #31, opened 2022-02-21
Improve testing of INC-137 & INC-224

ushkarev:


server/views/pages/incentives-table.njk
```diff
@@ -224,14 +224,11 @@
                 }
             }
 
-            ga('send', 'event', gaCategory, gaAction, '{{ locationPrefix }}')
+            if (typeof ga === typeof Function) ga('send', 'event', gaCategory, gaAction, '{{ locationPrefix }}')
```
xoen, line 228: I know it's kinda academic given in practice `ga()` would always be available but reason I had the `if` below was so that I didn't attach unnecessary event handlers (potentially slow) when nothing would happen in response to them.

But as I said, in our case this `if` move doesn't really change much ☺️

xoen (approved):

server/views/pages/incentives-table.njk
ushkarev: the GA integration test wouldn't work without this change. i had to move it otherwise the GA spy could not be inserted "soon enough" – essentially, the event listeners would never get added because `ga` doesn't exist as the page loads. there seems to be no way to get cypress to insert code _before_ the page

PR #37, opened 2022-02-24
INC-247: Mockup behaviour entry analytics page

ushkarev:


server/utils/format.ts
```diff
@@ -0,0 +1,24 @@
...
+        hour12: false,
+        minute: '2-digit',
+        timeZone: 'Europe/London',
+      })
+      .replace(',', '')
+  },
+
+  thousands(integer: number) {
+    if (notNumber(integer)) return '?'
+    return Math.round(integer).toLocaleString()
```
xoen, line 17: I wonder if given the name of the function we should be more explicit and specify the locale, e.g.

```
return Math.round(integer).toLocaleString('en-GB')
```

server/config.ts
```diff
@@ -98,6 +98,9 @@ export default {
   analytics: {
     googleAnalyticsId: get('GOOGLE_ANALYTICS_ID', ''),
   },
+  featureFlags: {
+    showAnalytics: get('SHOW_ANALYTICS', '') === 'true',
```
xoen, line 102: I like that this is in it's own "namespace" as we may add more feature flags.

Would be odd to prepend these (future) environment variables with something to group them together? e.g. `FF_SHOW_ANALYTICS`? Or maybe `FEATURE_SHOW_ANALYTICS`?

server/routes/testutils/appSetup.ts
```diff
@@ -91,6 +92,7 @@ function appSetup(production: boolean, testSession: Session, mockUserService: Us
   app.use('/select-location', selectLocationRoutes(standardRouter(mockUserService)))
   app.use('/incentive-summary/:locationPrefix', incentivesTableRoutes(standardRouter(mockUserService)))
   app.use('/prisoner-images/:imageId.jpeg', prisonerImagesRoutes(imageRouter()))
+  app.use('/analytics', analyticsRouter(standardRouter(mockUserService)))
```
xoen, line 95: Sorry, this is probably needs tweaking after the middleware refactoring, it would go under [`routes/all`](https://github.com/ministryofjustice/hmpps-incentives-ui/blob/main/server/routes/all.ts).

The good news is that you don't have to repeat this in two places ☺️

server/app.ts
```diff
@@ -41,6 +42,7 @@ export default function createApp(userService: UserService): express.Application
   app.use('/select-location', selectLocationRoutes(standardRouter(userService)))
   app.use('/incentive-summary/:locationPrefix', incentivesTableRoutes(standardRouter(userService)))
   app.use('/prisoner-images/:imageId.jpeg', prisonerImagesRoutes(imageRouter()))
+  app.use('/analytics', analyticsRouter(standardRouter(userService)))
```
xoen, line 45: I think this route should be added only if the feature is enabled.

I know the card would be hidden but potentially you could get there directly. In this case it probably doesn't matter much because there wouldn't be any harm, but I thought I'd mention.

server/routes/analyticsRouter.test.ts
```diff
@@ -0,0 +1,41 @@
+import type { Express } from 'express'
+import request from 'supertest'
+
+import { appWithAllRoutes } from './testutils/appSetup'
+
+let app: Express
+
+beforeEach(() => {
+  app = appWithAllRoutes({})
+  app.locals.featureFlags.showAnalytics = true
```
xoen, line 10: It's nice how this makes it easy to change the flag (and all locals in general) 👍

assets/sass/components/_chart-colours.scss
```diff
@@ -0,0 +1,41 @@
+.app-chart-colour--a {
+  background: govuk-tint(govuk-colour("blue"), 51%); // miro ~ 90aadc
+}
+
+.app-chart-colour--b {
+  background: govuk-tint(govuk-colour("dark-blue"), 18%); // miro ~ 2f5598
+}
```
xoen, line 7: Would be useful to have these 2 colours in variables like the `$heat-map-colour` one?

Also, maybe we'll use these for other meanings but it seems like at the moment these means "positive behaviours/case notes" and "negative behaviours/case notes". 
So maybe we could call them `-pos`/`-neg` or even `-positive`/`-negative`, just an idea.

server/views/pages/analyticsBehaviourEntries.njk
```diff
@@ -19,7 +19,84 @@
...
+                <br />
+                {{ row.entriesNegative | thousands}} <span class="govuk-visually-hidden">entries</span>
+              </td>
+              <td aria-hidden="true" class="app-chart-table__percentage-cell">
+              <span class="app-chart-table__axis-midpoint"></span>
+                <div id="c-pos-{{ loop.index }}" class="app-chart-table__percentage-cell-bar app-chart-colour--a">&nbsp;<br />&nbsp;</div>
+                <div id="c-neg-{{ loop.index }}" class="app-chart-table__percentage-cell-bar app-chart-colour--b">&nbsp;<br />&nbsp;</div>
+                <style nonce="{{ cspNonce }}">
+                  #c-pos-{{ loop.index }} { width:{{ percentagePositive }}; }
+                  #c-neg-{{ loop.index }} { width:{{ percentageNegative }}; }
```
xoen, line 77: 🧙

xoen (dismissed):
I left some comments/idea but LGTM ☺️

assets/sass/components/_chart-colours.scss
ushkarev: i didn't go for pos/neg because there will be other charts in future and they may use a larger palette. same with the classes: i'd rather they could be used in any chart rather than be tied to a word that doesn't then make sense.

as for using variables, that's a good idea but the colours themselves are very WIP at the moment and justin is looking into it. i basically tried to tint GDS colours to get as near to miro designs as possible.

i'm gonna leave their naming for now with the view to it all possibly changing anyway

server/views/pages/analyticsBehaviourEntries.njk
ushkarev: CSP does not allow us to use style attributes. ugh. makes charts rather hard to do!
there are some options…

assets/sass/components/_chart-colours.scss
xoen: I don't think it matters much at this stage as you said but I think I prefer the words even if we get a larger palette as they express the meaning of the colour rather than the...colour (e.g. "warning" > "yellow").

If a colour would be used in different contexts nothing wrong with having more than one variables with the same value if they refer to different meanings.

But yeah, don't worry about the names, it was a thought ☺️

server/app.ts
ushkarev: instead, i'll make it 404. otherwise tests won't work

server/routes/analyticsRouter.ts
```diff
@@ -1,10 +1,21 @@
...
-  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+  const featureGate = (handler: RequestHandler): RequestHandler => {
+    return (req: Request, res: Response, next: NextFunction): void => {
+      if (req.app.locals.featureFlags.showAnalytics) {
+        handler(req, res, next)
+      } else {
+        next(new NotFound())
+      }
+    }
+  }
```
xoen, line 16: Wouldn't be easier to add a similar `if` statement in the `routes/all.ts` file? effect should be the same right?

xoen (dismissed):

server/routes/analyticsRouter.ts
ushkarev: do you mean so that the analytics router isn't added to the parent? it wouldn't be possible to test the feature gate cos the router tree is only setup once, right?

helm_deploy/hmpps-incentives-ui/values.yaml
```diff
@@ -56,8 +56,6 @@ generic-service:
 
   allowlist:
     global-protect: "35.176.93.186/32"
-    moj-official-tgw-prod: "51.149.250.0/24"
```
xoen, line 58: well spotted! Maybe for the future we should aim at trying to have lists like this in alphabetical order as it should help spot things like this more easily ☺️

xoen (approved):

helm_deploy/hmpps-incentives-ui/values.yaml
ushkarev: i didn't compare the ranges themselves, however. it's possible they overlap somewhere

server/routes/analyticsRouter.ts
ushkarev: aah actually: conditionally adding this router in `routes/all.ts` doesn't work for a different reason. there's no way i can see to reach the `app.locals` object in that parent router without wrapping in a request handler that'd do basically the same thing as this.

i think i'll make this feature gate a shareable utility though

PR #50, opened 2022-02-25
INC-269: Compress images for improved loading

ushkarev:
The SVG logo sent with every request is shrunk by 6kb

Before:
Before

After:
After


server/views/partials/header.njk
```diff
@@ -4,9 +4,8 @@
   <div class="hmpps-header__container">
     <div class="hmpps-header__title">
       <a class="hmpps-header__link hmpps-header__title__organisation-name" href="{{ dpsHome }}">
-        <svg role="presentation" focusable="false" class="hmpps-header__logo" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 40 40" height="40" width="40">
```
ushkarev, line 7: GDS now recommend an `aria-hidden` attribute instead of a presentation role

Mjwillis (approved):
LGTM

PR #64, opened 2022-03-08
Tech improvements

ushkarev:


server/middleware/setUpWebSecurity.ts
```diff
@@ -29,6 +29,7 @@ export default function setUpWebSecurity(): Router {
           imgSrc: ["'self'", 'https://www.google-analytics.com'],
         },
       },
+      crossOriginEmbedderPolicy: true,
```
ushkarev, line 32: without this, remote jquery ui components do not load properly. i don't think this is the right fix, but it's what other projects are doing.

1) i don't think jquery ui is in fact used anywher
2) the proper solution is to load all front-end assets from the same origin

xoen: I thought jQuery UI was used by some of the govuk components, e.g. tabs? But I tried to comment the import and things seem to work fine?

1. If jQuery UI is not used, shouldn't we remove the import maybe?
2. true, there is a `jquery-ui-dist` node package or maybe we just download the files...

However, if we can be sure jQuery UI is not used probably remove is the best.

ushkarev: GDS design system does not use jquery or jquery ui. MoJ design system _might_ but i don't think it does.

xoen (approved):
This looks mostly good.

Not a fan of enabling `crossOriginEmbedderPolicy` but I'm sure neither you are - and there is a story to investigate removal of jQuery UI so 👌

PR #65, opened 2022-03-09
INC-329: Removed (unused) jQuery UI and simplified CSP

xoen:
jQuery UI was included in all pages (as it was part of the layout
template) but it doesn't seem to be used anywhere.

Even worse, the js/css wasn't served by the application itself which is
a potential security issue.

Remove these means we can remove the CSP rules to allow requests to
code.jquery.com. It also means we can remove the
`crossOriginEmbedderPolicy: true` option which was added to avoid the
related CSP error.


xoen:
Tested locally with browser console kept open to see if any errors would pop up. AFAICT after this change things seem to work as before (as expected)

ushkarev (approved):

PR #67, opened 2022-03-10
INC-343: Mockup protected characteristics graphs

ushkarev:


server/routes/analyticsRouter.ts
```diff
@@ -51,8 +51,11 @@ export default function routes(router: Router): Router {
       { text: 'Protected characteristics' }
     )
 
-    const prisonersByEthnicity = { dataSource: 'NOMIS', lastUpdated: new Date() }
-    const prisonersInAgeGroups = { dataSource: 'NOMIS', lastUpdated: new Date() }
+    const activeCaseLoad = res.locals.user.activeCaseload.id
+
+    const analyticsService = new AnalyticsService(urlForLocation)
+    const prisonersByEthnicity = await analyticsService.getIncentiveLevelsByEthnicity(activeCaseLoad)
+    const prisonersInAgeGroups = await analyticsService.getIncentiveLevelsByAgeGroup(activeCaseLoad)
```
xoen, line 58: Just a though so no need to change anything (especially given all is in flux), but it seems like these two could be done "concurrently" as the second call doesn't depend on the response of the first call.

Could be a good use case of `Promise.all`, e.g.

```javascript
const [prisonersByEthnicity, prisonersInAgeGroups] = await Promise.all([
    analyticsService.getIncentiveLevelsByEthnicity(activeCaseLoad),
    analyticsService.getIncentiveLevelsByAgeGroup(activeCaseLoad),
])
```

(or something like that - but as I said, just a thought, what you have is perfectly fine and clear)

ushkarev: you're right if the 2 data sets come from different files in S3. however, i'm pretty sure that they will actually be from one table grouped by different columns so it won't make a difference as there will only be one async call to S3

server/services/analyticsService.ts
```diff
@@ -141,4 +151,76 @@ export default class AnalyticsService {
...
+      ['Black or Black British', 8, 62, 46, 2],
+      ['Mixed', 2, 30, 40, 0],
+      ['Other', 1, 7, 8, 0],
+      ['White', 28, 615, 623, 2],
+    ]
+
+    const totals: number[] = []
+    for (let i = 0; i < levels.length; i += 1) {
+      totals.push(0)
+    }
```
xoen, line 172: I never tried it but apparently Array has a `fill()` method, something like:

```javascript
const totals: number[] = new Array(levels.length).fill(0)
```

It may read better and maybe even be more efficient...I'm just throwing it as it's a nice trick more than to say you should change it :)

ushkarev: yes, good idea. can i do that in the S3 PR? this file is changing a lot in the other branch and i'll have terrible merge conflicts as it is.

i'm coming across more and more places where i can generalise these functions

xoen (approved):

server/services/analyticsService.ts
xoen: Of course! As I said, was mainly just sharing a cool/concise way of doing it, don't have to change and it can certainly be done in a separate PR :)

PR #72, opened 2022-03-14
INC-289: Add S3 client

ushkarev:
• Adds a simple wrapper around AWS S3 SDK to list and get objects held in a predefined bucket. The client uses access credentials when provided, otherwise falls back to automatic methods of authentication – which is IRSA in the deployed application.
• Runs minio to mock S3 when running application or integration tests locally.
• Runs minio to mock S3 when running integration tests on CircleCI.

There are no test data files of interest yet, nor is the S3 client used in the application.


Mjwillis (approved):
LGTM

PR #75, opened 2022-03-15
INC-289: Analytics sample data & minor refactor

ushkarev:


server/services/analyticsService.ts
```diff
@@ -56,7 +65,25 @@ type PrisonersOnLevelsByAgeGroup = {
 }
 
 export default class AnalyticsService {
-  constructor(private readonly urlForLocation: (prison: string, location: string) => string) {}
+  constructor(
+    private readonly client: S3Client,
+    private readonly urlForLocation: (prison: string, location: string) => string
+  ) {}
+
+  async findTable<T extends Table>(tableName: string): Promise<{ table: T; date: Date; modified: Date }> {
```
xoen, line 73: I'd consider splitting this into two methods, a `findTable()` which returns the latest key for a table, and a `loadTable()` which reads the S3 object contents, parse it and mould it into the shape you expect from the outside.

Maybe it's overkill (it would mean two method invocations) but I like the idea of these methods doing only on thing. Just an idea :)

xoen (approved):

PR #83, opened 2022-03-18
INC-291: Investigate Redis/Sentry crash

xoen:
As a result of the investigation I've tweaked the code in few places:
- Sentry initialisation happens earlier in the application entrypoint
- Incentives table request handler doesn't create a new Redis client on each separate request anymore. This could be significant (however it's hard to check)
  - AFAICT use the same `HmppsAuthClient`/`TokenStore`/Redis client across all these requests should be fine because these classes don't really maintain any state (the state itself should be in the Redis server itself)
- attach Redis `'error'` event handler as soon as the client is created, this logs the error but it should prevent the Node process to exit
  - log message shows which Redis client was affected by the error, this is important because we have a few instances of Redis clients
- also, Sentry initialisation and Sentry request handler are now separate
- added a test endpoint which can be used to trigger the same error (behind feature flag, so disabled everywhere by default but easy to enable)

### Observations

I've tested this branch in `preprod` by building this docker image locally and pushing to ECR. 
Although it's not clear why these changes had a positive impact - it appears the application with these changes no longer crashes.

This compares positively with the behaviour we observed on main branch when Sentry was enabled: The application was throwing `The client is closed` errors (we think this is [caused by the session middleware not awaiting the Redis client `.connect()`](https://github.com/ministryofjustice/hmpps-incentives-ui/blob/015732a6b31edfc302ffc31fbfcd09f79693a29d/server/middleware/setUpWebSession.ts#L12-L13) - consistent to [`redis/node-redis`' docs](https://github.com/redis/node-redis/blob/master/docs/v3-to-v4.md#no-auto-connect)).
These errors were caught by Sentry and reported to Sentry.io but the node process didn't exit, which meant k8s wouldn't detect a problem and **not** restart the offending container - this would leave the application "up" but not working properly.

### Conclusion

This problem is very difficult to investigate and reproduce but it seems to be caused by the session middleware not waiting for the Redis connection to be established as it should.

Unfortunately use `await` is not trivial as everything would have to change to use `async function`s.

It seems this is a race condition which somehow is not manifesting itself with these changes. However we can't rule out that in the future this problem could reappear.

Having said that, these tweaks are reasonable and seem to have a positive impact which means we can use Sentry reporting. So we may decide to merge this and investigate further when/if the problem will ever reappear.


ushkarev (dismissed):
worth a try 🤷

ushkarev (approved):

PR #85, opened 2022-03-23
INC-381 & INC-388: Per-user & per-prison analytics feature gates

ushkarev:
Not only does the analytics feature need to be enabled for users to see that section, but also:
• their _active_ case load must be permitted
• their username must be permitted

… with both active case load and username restrictions allowing all or none or a specified set.


server/middleware/featureGate.test.ts
```diff
@@ -1,43 +1,95 @@
...
+
+/**
+ * Creates a mock response with given response locals
+ * NB: the send method will track calls
+ */
+function mockResponse(locals: Record<string, unknown> = {}): Response {
+  return {
+    locals,
+    send: jest.fn(),
+  } as unknown as Response
```
xoen, line 27: how does this `as unknown as Response` work?

ushkarev: `Response` is a huge and complicated type. the internet suggests that you mock and set the bits you need. `as Response` does not work because the type of this partial object is not compatible. so you need to force the type assertion. anything can be asserted as being `unknown` and anything can be made out of an `unknown` with a type assertion.

basically you gotta do this unless you want to implement 40 properties on this object 🤷

xoen: Oh I see now, you trick TS to think that's is `unknown` and then trick it again to think it's a `Response`: It's still missing many properties but now TS doesn't know 👌

server/middleware/featureGate.test.ts
```diff
@@ -1,43 +1,95 @@
...
+
+  it('calls handler when user’s active case load is included in specified prisons', () => {
+    const req = mockRequest()
+    const res = mockResponse({ user: { activeCaseload: { id: 'MDI' } } })
+    expectRequestHandlerToBeCalled(gatedHandler, req, res)
+  })
+
+  it('calls handler when user’s case loads overlap with specified prisons', () => {
+    const req = mockRequest()
+    const res = mockResponse({ user: { activeCaseload: { id: 'LEI' }, caseloads: [{ id: 'LEI' }, { id: 'BWI' }] } })
```
xoen: Maybe it's explained later in the diff but I'm confused by the fact this wouldn't receive a `404` response given the graph would still use data from `LEI` but that prison doesn't have the flag enabled (in this case it's enabled only for BWI/MDI).

server/middleware/featureGate.ts
```diff
@@ -1,9 +1,32 @@
...
+ * Wraps a request handler and returns 404 unless any of the user’s case loads appears in given prisons
+ */
+export function caseloadGate(prisons: string[], handler: RequestHandler): RequestHandler {
   return (req: Request, res: Response, next: NextFunction): void => {
-    if (req.app.locals.featureFlags[flag]) {
+    const activeCaseload: Caseload | undefined = res.locals?.user?.activeCaseload
+    const caseloads: Caseload[] = res.locals?.user?.caseloads ?? []
+    if (
+      (activeCaseload && prisons.includes(activeCaseload.id)) ||
+      caseloads.map(caseload => caseload.id).some(prison => prisons.includes(prison))
```
xoen, line 32: I think this middleware should only check for the active case load and ignore the other caseloads a user may have (maybe the name of the middleware should reflect this? e.g. `activeCaseLoadGate`? 🤷)

Assuming the handler uses the active case load to do something with it (e.g. [filter data/show a graph for the active case load prison](https://github.com/ministryofjustice/hmpps-incentives-ui/pull/85/commits/2646bb775b7ef619c56547e3c17146ba757e6d5e#diff-46a3cef6cd40be5b2c529ed431954c6e9b197b0218eed6ef43ddbf99b7286972R33)).

Is this to potentially avoid confusion for users with some case loads where analytics are shown but not in all?

server/middleware/featureGate.ts
```diff
@@ -37,3 +37,20 @@ export function caseloadGate(prisons: string[], handler: RequestHandler): Reques
     }
   }
 }
+
+/**
+ * Wraps a request handler and returns 404 unless the user’s *active* case load appears in given prisons
+ * If "*" appears in the prisons list, call is always forwarded to request handler
+ */
+export function activeCaseloadGate(prisons: string[], handler: RequestHandler): RequestHandler {
```
xoen, line 45: Oh I see, there are 2 middleware, one checks only the active case load, the other checks any user' case load - sorry didn't see this 👌

server/middleware/featureGate.ts
ushkarev: that's in exactly what i have in a subsequent commit! i thought it'd be more readable to have it in separate ones

server/middleware/featureGate.test.ts
xoen: I know see there are two separate middleware with different behaviour 👌

server/config.ts
```diff
@@ -120,6 +120,7 @@ export default {
     addTestErrorEndpoint: flag('FEATURE_ADD_TEST_ERROR_ENDPOINT', false),
     showAnalytics: flag('FEATURE_SHOW_ANALYTICS', false),
   },
+  prisonsWithAnalytics: (get('PRISONS_WITH_ANALYTICS', null, notRequiredInProduction) ?? '').split(','),
```
xoen, line 123: If you default to `''`, do you still need the `?? ''`, I wouldn't think so given environment variables should always be string or not set at all.

I wonder if that's equivalent to:

```
prisonsWithAnalytics: get('PRISONS_WITH_ANALYTICS', '', notRequiredInProduction).split(','),
```

🤔

ushkarev: the default from `get` is `null` when the env var is missing

xoen: That's very odd, by the look of `get()`, if you pass an empty string as fallback it should return it when the environment variable is not set 🤔 🤷

xoen (approved):

PR #86, opened 2022-03-23
INC-391: Use GDS copy for 404 and 500 pages

ushkarev:
Ref: [404](https://design-system.service.gov.uk/patterns/page-not-found-pages/) and [500](https://design-system.service.gov.uk/patterns/problem-with-the-service-pages/)

### DIFF
--- a/server/routes/testutils/appSetup.ts
+++ b/server/routes/testutils/appSetup.ts
@@ -1,5 +1,5 @@
-import express, { Express } from 'express'
-import { Cookie, Session, SessionData } from 'express-session'
+import express, { type Express, type Router } from 'express'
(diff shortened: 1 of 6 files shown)


server/routes/testutils/appSetup.ts
```diff
@@ -98,13 +108,15 @@ function appWithAllRoutes({
   production = false,
   testSession = makeTestSession(),
   mockUserService = new MockUserService(),
+  testRouter = undefined,
```
xoen, line 112: any reason for using `undefined` instead of `null` as default value?

xoen (dismissed):

xoen (approved):

PR #95, opened 2022-03-28
INC-390: Update analytics tile content to reflect current IA structure

ushkarev:


server/routes/analyticsRouter.test.ts
```diff
@@ -50,30 +50,30 @@ describe('Home page shows card linking to incentives analytics', () => {
   it('if feature is turned on', () => {
     return request(app)
       .get('/')
-      .expect(res => expect(res.text).toContain('Incentives data'))
+      .expect(res => expect(res.text).toContain('/analytics'))
```
ushkarev, line 53: look for URL rather than tile text so that it can vary without tests needing to be changed each time

PR #96, opened 2022-03-28
INC-385: Remove 'Days since last review' and 'Days on level' columns

xoen:
- added `hideDaysColumnsInIncentivesTable` feature flag which determines whether these two columns would be hidden (new behaviour) or not (legacy behaviour)
- flag is off by default, meaning columns are shown by default as per original behaviour
- table headers were small when we had all the columns but their original font size is restored when these 2 columns are removed (as per prototype)
- Incentive table page heading copy updated as per [ticket INC-390](https://dsdmoj.atlassian.net/browse/INC-390) - copy changes only when flag is on
- A message is shown on the Incentive table page to inform the users that the columns were removed as per [ticket INC-384](https://dsdmoj.atlassian.net/browse/INC-384) - message shown only when flag is on
- home page Incentive table tile copy updated as per [ticket INC-390](https://dsdmoj.atlassian.net/browse/INC-390) - copy changes only when flag is on


server/routes/homeRouter.test.ts
```diff
@@ -0,0 +1,48 @@
+import type { Express } from 'express'
+import request from 'supertest'
+
+import { appWithAllRoutes } from './testutils/appSetup'
+
+let app: Express
+
+beforeEach(() => {
+  jest.clearAllMocks()
```
ushkarev, line 9: i suspect you don't need this, but no harm in it i guess

helm_deploy/values-prod.yaml
```diff
@@ -16,6 +16,7 @@ generic-service:
     HMPPS_PRISON_API_URL: "https://api.prison.service.justice.gov.uk"
     NOMIS_USER_ROLES_API_URL: "https://nomis-user.aks-live-1.studio-hosting.service.justice.gov.uk"
     TOKEN_VERIFICATION_API_URL: "https://token-verification-api.prison.service.justice.gov.uk"
+    FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE: "false"
```
Mjwillis, line 19: why false in prod? Is to validate first?

Mjwillis (approved):
LGTM

ushkarev (approved):

helm_deploy/values-prod.yaml
xoen: Igor added some utility to read feature flags values and [it would default to false anyway](https://github.com/ministryofjustice/hmpps-incentives-ui/compare/ag--INC-385-hide-days-columns?expand=1#diff-b5e32f68bcc433506f573cc64aae69018f0a1caa3822686806a704ac1c8e906cR122) but I thought it would be good to be explicit and spell it out even if the effect is the same.

server/routes/homeRouter.test.ts
xoen: Yeah I was copying this from the other test where you check the analytics tiles. Originally I was tempted to move those tests into this file (they mock S3 things), but then I decided to not do so and keep this changeset focused on the table stuff.

PR #98, opened 2022-03-28
INC-412 & INC-413: Add feedback forms to analytics charts

ushkarev:
To get feedback from users on each individual chart, a form (hidden in a details element) has been added above each chart along with the guidance boxes. Successful submissions are posted as tickets in Zendesk, appropriately tagged so that they can be easily counted and collated by prison.

Sample:
chart-feedback


Mjwillis (approved):
LGTM

PR #107, opened 2022-03-30
INC-437: Make form instances usable when data was not submitted

ushkarev:
…so that templates can always rely on their existence and their methods. Previously, the only situation when a form variable was passed into the page template was when the form was submitted but did not pass validation. This made templates unintuitive (most of the time the form variable did not exist).

This paves the way to allow forms to preset initial values before a form submission.

### DIFF
--- a/server/routes/forms/forms.ts
+++ b/server/routes/forms/forms.ts
@@ -1,26 +1,91 @@
-export default abstract class Form<Data extends { formId: string }> {
-  // posted form data; NB: it may be transformed in place by validation
-  data: Partial<Data>
+export interface BaseFormData {
+  /**
+   * Unique identifier to distinguish multiple forms on one page
+   */
+  formId: string
+}
 
-  // fields with error messages; for use with error message components
-  fieldErrors: Partial<Record<keyof Data, string>>
+/**
+ * Base form providing simple validation extension points and per-field error messages
+ */
+export default abstract class Form<Data extends BaseFormData> {
+  /**
+   * Posted form data, undefined indicates that the form has not been submitted/POSTed
+   * NB: it may be transformed in place by validation
+   */
+  protected data?: Partial<Data>
 
-  constructor(data: Partial<Data>) {
-    this.data = data
+  /**
+   * Holds per-field error messages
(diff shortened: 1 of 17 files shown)


server/routes/forms/forms.ts
```diff
@@ -1,26 +1,91 @@
...
+    /**
+     * Unique identifier to distinguish multiple forms on one page
+     */
+    readonly formId: string
+  ) {
     this.fieldErrors = {}
+  }
+
+  public toString(): string {
+    return `[Form formId=${this.formId} submitted=${this.submitted} hasErrors=${this.hasErrors}]`
```
xoen, line 33: Maybe I'm missing some JS subtlety here but shouldn't `this.submitted()`/`this.hasErrors()` be invoked, e.g. have `()` after their names?

I don't imagine the string interpolation would do that automatically 🤔 ?

ushkarev: they're both getter methods, so `form.submitted` calls the method

xoen: Cheers, I missed that! Nice ☺️

server/views/partials/analytics/chartFeedbackForm.njk
```diff
@@ -2,14 +2,13 @@
...
+{# Required variables: chartFeedbackForm, csrfToken #}
 
-{% macro commentTextarea(graphId, id) %}
+{% macro commentTextarea(chartFeedbackForm, id) %}
   {{ govukTextarea({
-    id: graphId + "-" + id + "Comments",
+    id: chartFeedbackForm.formId + "-" + id + "Comments",
     name: id + "Comments",
-    value: form.data[id + "Comments"] | default(""),
+    value: chartFeedbackForm.getField(id + "Comments").value,
```
xoen, line 11: Just double checking, when the value is not in the form data `.value` would be `undefined`, but this is fine because `govukTextarea()` would just ignore it right? It's not gonna be converted to `'undefined'` right?

xoen (approved):
Nice, looks cleaner 👍

PR #113, opened 2022-04-01
Test analytics location and characteristic sorting more thoroughly

ushkarev:
…because "Unknown" should appear last. Though it does not matter yet as those data points do not yet show in any charts.


server/services/analyticsService.test.ts
```diff
@@ -132,12 +132,16 @@ describe('AnalyticsService', () => {
   describe('comparators and filters', () => {
     describe.each([
       { a: { location: 'All' }, b: { location: '1' }, expected: -1 },
+      { a: { location: 'Unkown' }, b: { location: '1' }, expected: 1 },
```
xoen, line 135: I think the spelling is missing an 'n', e.g. `Unkown` => `Unknown`.

xoen (approved):

PR #114, opened 2022-04-01
INC-462: Bucket tools

ushkarev:
These are simple command line tools, run manually on pods (`kubectl exec …`), for testing/debugging purposes rather than for use in the website/application.

### DIFF
--- a/bin/download-source-tables.ts
+++ b/bin/download-source-tables.ts
@@ -1,0 +1,35 @@
+#!/usr/bin/env npx ts-node
+import fs from 'fs'
+import path from 'path'
+
+import config from '../server/config'
+import S3Client from '../server/data/s3Client'
+import { TableType } from '../server/services/analyticsServiceTypes'
+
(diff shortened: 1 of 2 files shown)


bin/download-source-tables.ts
```diff
@@ -0,0 +1,35 @@
...
+  const scriptName = path.basename(process.argv[1])
+  const tableTypeOptions = tableTypes.map(t => `"${t}"`).join(' | ')
+  process.stderr.write(
+    `Usage: ./${scriptName} [${tableTypeOptions}] [YYYY-MM-DD] [path to save to or '-' for stdout]\n` +
+      `NB: Only works on pods within Cloud Platform\n`
+  )
+  process.exit(1)
+}
+
+const [tableTypeKey, date, savePath] = process.argv.slice(2, 5)
```
ushkarev, line 21: i think i might enhance this with `yargs` or similar in future, because this is more error prone

xoen (approved):
LGTM

PR #117, opened 2022-04-06
INC-463: Copy change to 'analytics' tile heading

xoen:
Linsey pointed out the 's' in the word 'Incentives' also needs to go:

> It should say:
> Incentive level and behaviour entry data.


neilmendum (approved):
LGTM

PR #121, opened 2022-04-12
Cache Analytics' stitched tables values

xoen:
In-memory cache for Analytics' stiched tables.

The stitched tables are cached based on the source S3 table name
and the columns selected. Adding more graphs using the same data but
different columns should not cause problems in the future.

This cache will be shared across all the requests served from the same
Node process. Size of biggest stitched table is currently ~30MB which
is small enough to be kept around the process, this means we avoid the
overhead of talking to Redis.
Further clear this cache would be as simple as restart the k8s pods (if
we need to do so).

NOTE: We only cache the matrix result with the filtered columns, the
request handlers will still do the filtering by prison and other aggregations.
This is by design because, the filtering/aggregation step is relatively
quick and to cache those very specific/small artefacts would probably mean
a lot of cache misses (as cache would be only useful on a prison-by-prison
basis).


xoen:
(rebased)

xoen:
(rebased after new pc graphs were added in `main`)

ushkarev (approved):
maybe should mention in team chat about not displaying latest charts for up to 2 hours…? i know we had previously decided that "a day old" is not too stale, but *if* ppl feel strongly about showing the latest _possible_ then we could either:
• always list files in s3 and use file name as part of cacheKey (@neilmendum's idea)
• have some kind of background process to list files in s3 and clear the whole cache when necessary

server/config.ts
```diff
@@ -133,6 +133,7 @@ export default {
     showAnalytics: flag('FEATURE_SHOW_ANALYTICS', false),
     showPcAnalytics: flag('FEATURE_SHOW_PC_ANALYTICS', false),
   },
+  analyticsCacheExpiryMinutes: Number(get('ANALYTICS_CACHE_EXPIRY_MINUTES', 120)),
```
neilmendum, line 136: Thinking there are a couple of scenarios where having a cache expiry impacts the user experience because they would need to wait 2 hours for the data to reflect the latest extract
- Daily extract is typically produced around 2pm and not visible until 4pm
- If there was an issue with the extract and the analytics team needed to the re-run the job then again there would be a delay in seeing the latest data
  - This is a failure scenario rather than BAU so perhaps in this case we would manually restart the pods to clear the caches, but that would add operational overhead

It's probably worth mentioning this to the rest of the squad so they're aware of the impact?

One option is to
1. Include `modified` in `cacheKey`
2. Check S3 for every request and find compare the latest `modified` vale and compare it with the cache

server/services/analyticsService.ts
```diff
@@ -335,3 +351,107 @@ export function compareCharacteristics(
...
+   * the source table name and the columns (e.g. 'incentives_latest_narrow,prison,wing').
+   *
+   * The values contains the `expiresAt` date determining whether the value
+   * is expired or fresh.
+   *
+   * The `value` property contains the actual cached value, including
+   * the `stichedTable` matrix and the `date`/`modified` dates.
+   */
+  private static cache: Record<
+    string, // cacheKey [S3 object key, ...columnsToStitch].join()
```
neilmendum, line 387: Should this comment be
`// cacheKey [TableType, ...columnsToStitch].join()`
?

server/config.ts
xoen: Bear in mind that access to data is kinda random so the new data not being visible until 4pm is a **worst** case scenario rather than always happening (if previous value happened to be cached just 1 minute before new data, e.g. 13:59pm).

It's a cache so it's to avoid an expensive operation but it means data may be a bit out of date. Hence I wasn't too aggressive with the time this was cached. We could reduce this to 1h and maybe that's good enough.

I don't think option 1 is possible because `modified` is not known by the client and we cannot include it in the `cacheKey`.

Second option means we have to make an HTTP request to S3 every time to check if the cache is fresh (bear in mind in some pages we show 2-4 graphs). We _could_ do it, but bear in mind that currently `findTable()` also gets the data so to change the logic this will have to be split, not sure how easy that is 🤔 (at the very least there are quite a few tests to update too)

The simplest thing would be to maybe reduce the cache validity to 1h and accept that "old" data could be cached/shown for up to ~1h (worst case scenario, I'd imagine on average be ~half of that).

The alternative is more involved:
1. split `AnalyticsService.findTable()` into 2 methods
  1a. one that returns that returns the latest key/metadata
  2a. one that gets the actual content
2. update cache logic to always check latest modified and if current value has a lower modified update the cache (effectively remove the existing logic/setting to generate/check expiry time of cached value)

server/services/analyticsService.ts
xoen: Comment updated ☺️

ushkarev (dismissed):

server/config.ts
neilmendum: That's a good point - my comment was focusing on the worst case scenario 😄 

If we aren't too concerned with potentially showing out of date data then keeping the cache logic simple may be a better option. Probably helpful getting some input from rest of the squad

xoen: I'll mention this to others in the team. I wanna experiment a bit with refactoring the `findTable()` method and the caching logic (to compare new modified vs cached modified).

xoen:
FYI: I've added 3 commits to play with the idea of checking the `modified` date coming from S3 to determine whether the cached data is still fresh.

xoen:
(rebased)

server/services/analyticsService.ts
```diff
@@ -35,18 +34,35 @@ export default class AnalyticsService {
...
     const date = new Date(key.slice(key.length - 15, key.length - 5))
+
+    return { key, date, modified }
+  }
+
+  async getTable<T extends Table>(key: string): Promise<T> {
     const object = await this.client.getObject(key)
     try {
       const table = JSON.parse(object) as T
-      logger.info(`Found latest "${tableType}" table: ${key} (modified ${modified.toISOString()})`)
```
ushkarev, line 41: i'd still keep an info message here, e.g. "Downloaded table …"

xoen: FYI: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/121/commits/cbf1725aaebd098cbc31f17fbcafce4e5d2a1971

ushkarev (approved):

neilmendum (approved):
LGTM

PR #130, opened 2022-04-22
INC-490: Add sexual orientation protected characteristic chart

ushkarev:
(none of the protected characteristics charts show in production yet)


neilmendum (approved):
LGTM

PR #132, opened 2022-04-25
INC-498: Use consistent colour palette for analytics charts

ushkarev:
Currently, when a prison has nobody on “Basic” the colour palette shifts so that “Standard” uses the first / light blue colour.
It’s been suggested that levels should use the same colours even when some levels are missing.

### DIFF
--- a/server/utils/analytics.test.ts
+++ b/server/utils/analytics.test.ts
@@ -1,0 +1,56 @@
+import { Colour, makeChartPalette } from './analytics'
+
+describe.each([
+  [
+    'with typical incentive levels',
+    ['Basic', 'Standard', 'Enhanced'],
+    ['app-chart-colour--a', 'app-chart-colour--b', 'app-chart-colour--c'],
+  ],
+  [
+    'with extended incentive levels',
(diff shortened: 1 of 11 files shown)


server/utils/analytics.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  [
+    ['Basic', 'Standard', 'Enhanced', 'Super enhanced'],
+    ['app-chart-colour--a', 'app-chart-colour--b', 'app-chart-colour--c', 'app-chart-colour--d'],
+  ],
+  [
+    ['Basic', 'Standard', 'Enhanced', 'Enhanced 2', 'Entry'],
+    ['app-chart-colour--a', 'app-chart-colour--b', 'app-chart-colour--c', 'app-chart-colour--d', 'app-chart-colour--e'],
+  ],
+  [
+    ['Basic', 'Standard', 'Enhanced', 'Enhanced 2', 'A', 'B'],
```
neilmendum, line 29: Is this test validating that 2 unexpected column names are defaulted to `app-chart-colour--e`?

ushkarev: i'll add explanations…

PR #139, opened 2022-04-29
INC-507: Rewrite analytics charts using macros

ushkarev:
…instead of layers of inherited templates in order to make required parameters really explicit (there are many!)


neilmendum (approved):
LGTM

server/views/partials/analytics/chartBaseHeatmap.njk
```diff
@@ -19,6 +12,44 @@ Required variables (in addition to base template):
...
+  {{- column -}}
+{% endmacro %}
+
+{% macro rowTotalValue(total, rowIndex) %}
+  {{- total | thousands -}}
+{% endmacro %}
+
+{% macro rowValue(value, column, columnIndex) %}
+  {{- value | thousands -}}
+{% endmacro %}
```
xoen, line 25: `columnIndex`, `rowIndex` and `column` are not used in these macros right? I'm guessing in other versions of these macros they're used? Or maybe some of these are not used anymore?

May be worth marking them as "unused" by prefixing them with an underscore, e.g. `_columnIndex` - assuming I'm not missing something here.

server/views/partials/analytics/chartBaseStackedPercentageBars.njk
```diff
@@ -1,15 +1,46 @@
...
+  guidance=guidance,
+  graphId=graphId,
+  graphGoogleAnalyticsCategory=graphGoogleAnalyticsCategory,
+  activeCaseload=activeCaseload,
+  form=form,
+  csrfToken=csrfToken,
+  report=report
+) %}
+
 <div class="govuk-!-margin-top-8">
```
xoen, line 44: I wouldn't be against indenting the blocks between `{% call ...%}` and `{% endcall %}`.

ushkarev: there is a lot of indenting i want to do, but didn't in this PR cos then every line of the diff would look modified

xoen (approved):

server/views/partials/analytics/chartBaseHeatmap.njk
ushkarev: hmmmm that wouldn't work. i call the macros with keyword arguments (as opposed to positional arguments) so cannot rename them! and i do that because 1) at the call site, you can see which argument is which and 2) order doesn't matter.

plus these ones are so short that you can see what they use and what they don't; this macro definition is the contract that compatible macros must obey.

PR #142, opened 2022-05-04
INC-535: Update tile content to differentiate sections more clearly

ushkarev:


neilmendum (approved):
LGTM

xoen (approved):

server/views/pages/home.njk
```diff
@@ -13,17 +13,15 @@
   <ul class="govuk-grid-row card-group">
     <li class="govuk-grid-column-one-third card-group__item">
       {% if featureFlags.hideDaysColumnsInIncentivesTable %}
-        {% set cardHeading = 'View incentive levels and behaviour entries' %}
-        {% set cardDescription = 'See incentive levels and behaviour entries by residential location.' %}
+        {% set cardDescription = 'See incentive levels and behaviour entries for the prison population, by residential location.' %}
       {% else %}
-        {% set cardHeading = 'View review dates and incentive information' %}
         {% set cardDescription = 'See review dates, incentive levels and behaviour entries by residential location.' %}
```
xoen, line 18: Shouldn't we update the alternative version in case we don't hide the 'Days since last review'/'Days on levels' columns (e.g. `featureFlags.hideDaysColumnsInIncentivesTable === false`)?

I'm guessing it would be very similar but mention 'review dates' (e.g. `See review dates, incentive levels and behaviour entries for the prison population, by residential location.`).

Alternatively we don't care about mentioning 'review dates' and just always show the updated version (and we could get rid of the `if` and `cardDescription` variable.

Something to maybe check with Linsey?

xoen (approved):
LGTM, left a comment about the 'Incentives table' card which is still displaying different card descriptions for when the `hideDaysColumnsInIncentivesTable` feature flag is on/off.

server/views/pages/home.njk
ushkarev: the title content is clearly gonna change several times before the days-on-level comes back. i don't see it as a problem and this "fallback" copy is good enough since nobody will ever see it!

ushkarev: if you want to entirely remove the days-on-level stuff then i figure that ought to be a separate PR

PR #143, opened 2022-05-04
Send GA events also to Google Analytics 4 (GA4)

xoen:
Before this we were only sending custom events to Universal Analytics (UA).

UA will stop working at some point next year so in order to migrate to GA4 we'll need to also send events to it, this will mean that we'll eventually have enough data to stop using UA.

GA4 events are sent using `gtag()` function. GA4 events are more flexible and don't have the same rigid structure that UA events had (all events have same category/event/action properties).

Having said that, because of the way we display heterogeneous events from different pages into a single table in Google Data Studio and for simplicity sake, I've retained the same structure that old UA events had.

All of the custom events will have the same name (`incentives_event`) and they'll have the `category`/`action`/`label` properties that the UA events had.

### DIFF
--- a/assets/.eslintrc.json
+++ b/assets/.eslintrc.json
@@ -8,6 +8,7 @@
   },
   "globals": {
     // Google Analytics
-    "ga": true
+    "ga": true,
+    "gtag": true
   }
 }
--- a/integration_tests/integration/analytics.spec.ts
+++ b/integration_tests/integration/analytics.spec.ts
@@ -53,42 +53,50 @@ context('Analytics section', () => {
       const gaSpy = new GoogleAnalyticsSpy()
       gaSpy.install()
 
-      page.entriesByLocationGuidance
-        .click()
-        .then(() =>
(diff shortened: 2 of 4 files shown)


server/views/partials/layout.njk
```diff
@@ -16,6 +17,21 @@
...
+  {% if googleAnalyticsGa4Id %}
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id={{ googleAnalyticsGa4Id }}"></script>
+    <script nonce="{{ cspNonce }}">
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+
+      gtag('config', '{{ googleAnalyticsGa4Id }}', {
+        active_case_load: '{{ user.activeCaseload.id | default('') }}',
```
xoen, line 30: This property will be automagically added to all events by GA4, similarly to what done previously in UA via a custom dimension.

PR #145, opened 2022-05-10
INC-544: Removed some analytics feature flags/settings

xoen:
Removed the following feature flags because they're always enabled in production:
- `FEATURE_SHOW_ANALYTICS`
- `FEATURE_SHOW_PC_ANALYTICS`
- `PRISONS_WITH_ANALYTICS`
- `USERNAMES_WITH_ANALYTICS`


helm_deploy/values-dev.yaml
```diff
@@ -23,11 +23,7 @@ generic-service:
     TOKEN_VERIFICATION_API_URL: "https://token-verification-api-dev.prison.service.justice.gov.uk"
     PHASE_NAME: "DEV"
     FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE: "true"
-    FEATURE_SHOW_ANALYTICS: "true"
-    FEATURE_SHOW_PC_ANALYTICS: "true"
     FEATURE_SHOW_ANALYTICS_TRENDS: "true"
-    PRISONS_WITH_ANALYTICS: "BWI,MDI,WRI"  # sample data only exists for these 3
```
ushkarev, line 29: this reminds me that data is only available in these 3 prisons locally, on `dev` and `preprod`. removing this feature gate will show a hopefully-useful message rather than a 404

xoen: Don't know if it's worth doing anything special about it or not 🤔 .

AFAICT we'd get the same error message saying data is not available for that prison. Maybe that's enough?

server/views/partials/analytics/layout.njk
```diff
@@ -19,8 +19,7 @@
       text: "Protected characteristics",
       href: "/analytics/protected-characteristics",
       active: graphPageId == "protected-characteristics",
-      attributes: { "data-qa": "protected-characteristics" },
-      itemToRemove: not featureFlags.showPcAnalytics
+      attributes: { "data-qa": "protected-characteristics" }
     }
   ] | rejectattr('itemToRemove') %}
```
ushkarev, line 24: don't need to filter the list anymore. you could even pass the list of items straight into the sub-nav macro

xoen: Fair enough 👌

ushkarev (approved):

PR #146, opened 2022-05-10
Ensure that all Google Analytics globals are defined in integration tests

ushkarev:


xoen (approved):
👍 and thanks Igor ☺️ !!

PR #147, opened 2022-05-10
INC-539: Prepare for using real data in analytics trends charts

ushkarev:
• added sample trends data
• factored out utility functions into a separate file
• made S3 mocking detect which table is being loaded – so that one page / route can present charts from multiple source tables
• added utility functions to ensure trends charts will show precisely the last 12 months and only them


bin/truncate-source-table.ts
```diff
@@ -37,6 +37,21 @@ if (chosenTableType === 'behaviourEntries') {
     'sex_orientation',
   ])
   rowFilter = (rowIndex, table) => characteristicsToKeep.has(table.characteristic[rowIndex] as string)
+} else if (chosenTableType === 'trends') {
```
xoen, line 40: May be worth using JS' `switch` now that we have 3 (four really) branches here.

server/services/analyticsServiceUtils.ts
```diff
@@ -36,6 +36,61 @@ export function mapRowsAndSumTotals<
...
+  const oldestMonth = yearAgo.getMonth()
+
+  return rows.filter(({ month: rowMonth }) => {
+    const year = rowMonth.getFullYear()
+    const month = rowMonth.getMonth()
+    return (
+      (year > oldestYear && year < latestYear) ||
+      (year === oldestYear && month >= oldestMonth) ||
+      (year === latestYear && month <= latestMonth)
+    )
```
xoen, line 90: I'm struggling to follow this a bit tbh. I guess we rely on the fact we don't have future months (e.g. or `(year > oldestYear && year < latestYear)` would mean something like July 2023 would be in rows).

Edge case aside, I'm wondering if it would be easier to use `Date` comparisons? e.g. _something like_:

```javascript
  const latestMonth = new Date()
  latestMonth.setDate(1)
  latestMonth.setHours(12, 0, 0, 0)
  latestMonth.setMonth(latestMonth.getMonth() - 1)

  const earliestMonth = new Date()
  earliestMonth.setDate(1)
  earliestMonth.setHours(12, 0, 0, 0)
  earliestMonth.setMonth(earliestMonth.getMonth() - 12)

  return rows.filter(({ month: rowMonth }) => {
    return rowMonth >= earliestMonth && rowMonth <= latestMonth
  })
```

or something like that.

xoen (approved):

server/services/analyticsServiceUtils.ts
ushkarev: i don't think date comparisons would be a good idea because of time zones. perhaps i could map the YEAR-MONTH strings into `Date`s _after_ fixing the timeline.

PR #162, opened 2022-05-12
INC-544: dropdown selector to navigate between protected characteristics pages

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-544

### DIFF
--- a/server/views/pages/analytics/protectedCharacteristicTemplate.njk
+++ b/server/views/pages/analytics/protectedCharacteristicTemplate.njk
@@ -1,6 +1,3 @@
-{% from "govuk/components/button/macro.njk" import govukButton %}
-{% from "govuk/components/select/macro.njk" import govukSelect %}
-
 {% extends "../../partials/analytics/layout.njk" %}
 
 {% from "../../partials/analytics/chartPrisonersOnLevels.njk" import chartPrisonersOnLevels %}
@@ -30,16 +27,16 @@
     <div class="govuk-grid-column-two-thirds">
       <form id="form-select-characteristic" method="get" novalidate>
         <div class="govuk-form-group">
-          <label class="govuk-label" for="sort">
+          <label class="govuk-label" for="characteristic">
             Select a protected characteristic
           </label>
           <select class="govuk-select" id="characteristic" name="characteristic">
               {%- for option in characteristicOptions -%}
-                <option value="{{ option.value }}" {%- if option.selected -%}selected{%- endif -%}>{{ option.label }}</option>
+                <option value="{{ option.value }}" {% if option.selected %}selected{% endif %}>{{ option.label }}</option>
               {%- endfor %}
           </select>
 
-          <button class="govuk-button govuk-!-margin-left-2" data-module="govuk-button">
+          <button class="govuk-button govuk-!-margin-left-2" data-module="govuk-button" data-prevent-double-click="true">
             Show results
           </button>
         </div>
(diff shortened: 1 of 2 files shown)


server/views/pages/analytics/protectedCharacteristicTemplate.njk
```diff
@@ -1,3 +1,6 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
```
ushkarev, line 1: these 2 imports are not used

server/views/pages/analytics/protectedCharacteristicTemplate.njk
```diff
@@ -23,6 +26,27 @@
     </div>
   </div>
 
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <form id="form-select-characteristic" method="get" novalidate>
+        <div class="govuk-form-group">
+          <label class="govuk-label" for="sort">
```
ushkarev, line 33: the label is not linked correctly to the select element.
the GDS form component templates are quite fiddly, so generally it's a good idea to use their macros so all the ids and names match up. plus, if GDS update them, we automatically benefit

server/views/pages/analytics/protectedCharacteristicTemplate.njk
```diff
@@ -23,6 +26,27 @@
...
+            Select a protected characteristic
+          </label>
+          <select class="govuk-select" id="characteristic" name="characteristic">
+              {%- for option in characteristicOptions -%}
+                <option value="{{ option.value }}" {%- if option.selected -%}selected{%- endif -%}>{{ option.label }}</option>
```
ushkarev, line 38: you _don't_ want to trim whitespace around the `selected` attribute otherwise it'll run together with the `value` attribute:

`<option value="age"selected>Age</option>`

server/views/pages/analytics/protectedCharacteristicTemplate.njk
```diff
@@ -23,6 +26,27 @@
...
+              {%- endfor %}
+          </select>
+
+          <button class="govuk-button govuk-!-margin-left-2" data-module="govuk-button">
```
ushkarev, line 42: i'm on the fence about preventing double-clicks… the page is not exactly quick to load. happy to leave it for now :)

ushkarev (commented):
which layout option did justin prefer in the end?
1) simple boxes floated left, all of natural size
2) button being inside the form group or bleeding into it using a negative margin
3) as in miro, the select box is deliberately made wider

Untitled drawing

i also think this is a good point to start adding a new test suite for this page

xoen:
> which layout option did justin prefer in the end?

@ushkarev , we had a chat with Justin and the design in the prototype is the correct one (I think that's option 2)

server/views/pages/analytics/protectedCharacteristicTemplate.njk
xoen: I'm sorting the wrong `for` attribute, well spotted ☺️ 

Regarding the use of the macros, agreed.  Unfortunately, given:
- the desired layout is the one where button is just next to select
- govuk's select macro wraps the select in a `.govuk-form-group` div

This is why we went the HTML route (given is relatively straightforward markup) because this way we can wrap both the select and the button elements in the same `.govuk-form-group` div.

server/views/pages/analytics/protectedCharacteristicTemplate.njk
xoen: Yeah, when we were working on this we didn't add it because it seemed overkill given it's a `GET` request.
But tbh it doesn't harm to have it so we're adding 👌

ushkarev (approved):

PR #164, opened 2022-05-13
INC-563: Release trends charts for incentive levels and behaviour entries

ushkarev:


xoen (dismissed):

xoen:
@ushkarev uh, diff looked fine but some tests are failing 🤔

ushkarev:
@xoen it was because of mixed `Report` types being passed through the `transformAnalyticsError` decorator… typescript was being a bit funny here, it compiled and passed tests completely fine on my machine. i've made the decorator less strict/fussy about what it passes through

neilmendum (approved):

PR #170, opened 2022-05-16
INC-544: Protected Characteristics tests

xoen:


integration_tests/integration/analytics.spec.ts
```diff
@@ -435,112 +455,96 @@ context('Analytics section', () => {
...
-  feedbackBoxes.forEach(([feedbackBox, feedbackForm]) => {
+  feedbackBoxes.forEach(graphId => {
     let page = Page.verifyOnPage(pageClass)
 
     // open feedback box and select "yes"
-    page[feedbackBox].click()
-    page[feedbackForm].find('[name=chartUseful][value=yes]').click()
-    page[feedbackForm].submit()
+    page.getGraphFeedback(graphId).click()
+    page.getGraphFeedbackForm(graphId).find('[name=chartUseful][value=yes]').click()
```
neilmendum, line 546: We refactored these 2 methods to (hopefully) simplify the calling code and reduce the number of accessor methods

integration_tests/integration/analytics.spec.ts
```diff
@@ -316,10 +373,9 @@ context('Analytics section', () => {
     it('users see analytics', () => {
```
neilmendum, line 373: Decided to only test the default protected characteristics page (age) because the other characteristics do not have any specific logic in the template

neilmendum, line 373: Decided to only test the default protected characteristics page (age) because the other characteristics do not have any specific logic in the template

integration_tests/integration/analytics/analyticsBehaviourEntries.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+    cy.task('stubAuthUser')
+    cy.task('stubCreateZendeskTicket')
+
+    cy.signIn()
+
+    const homePage = Page.verifyOnPage(HomePage)
+    homePage.viewAnalyticsLink().click()
+  })
+
+  it('has feedback banner', () => {
```
ushkarev, line 21: this test is repeated identically across the 3 analytics test suites. instead i'd move it into the body of the suite after the correct nav item has been clicked. that way each page will check for the banner rather than just the incentive levels page 3 times.

integration_tests/integration/analytics/utils.ts
```diff
@@ -0,0 +1,73 @@
+import Page, { PageElement } from '../../pages/page'
+import { AnalyticsPage } from '../../pages/analytics'
+
+export function testValidFeedbackSubmission<PageClass extends AnalyticsPage>(
+  pageClass: new () => PageClass,
+  feedbackBoxes
```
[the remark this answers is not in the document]
xoen: Yep, makes totally sense, we'll change 👌 (here and in the other function).

integration_tests/integration/analytics/analyticsBehaviourEntries.spec.ts
xoen: We've moved things around a bit as discussed this morning, hopefully this is similar to what you had in mind 👌 

https://github.com/ministryofjustice/hmpps-incentives-ui/pull/170/commits/3ae0a1c599122580be7152ae413d75994fff85e3

integration_tests/integration/analytics/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+      expect($divs).to.have.lengthOf(1)
+      expect($divs[0]).to.contain('Your feedback has been submitted')
+    })
+    page.errorSummary.should('not.exist')
+  })
+}
+
+export function testInvalidFeedbackSubmission<PageClass extends AnalyticsPage>(
+  pageClass: new () => PageClass,
+  feedbackBoxes
```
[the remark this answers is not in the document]
xoen: Done here (hopefully it's what you meant): https://github.com/ministryofjustice/hmpps-incentives-ui/pull/170/commits/6e4983d2f93b8cb9f008be8643c955e0747bfa03

PR #173, opened 2022-05-19
INC-544: remove feature flag and old template

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-544


server/routes/analyticsRouter.ts
```diff
@@ -235,54 +235,6 @@ export default function routes(router: Router): Router {
...
-    'incentive-levels-by-religion',
-    'incentive-levels-by-disability',
-    'incentive-levels-by-sexual-orientation',
-  ]
-  routeWithFeedback('/protected-characteristics', protectedCharacteristicsGraphIds, async (req, res, next) => {
-    res.locals.breadcrumbs.addItems({ text: 'Protected characteristics' })
-
-    // Redirect to new Protected characteristics page when feature flag is on
-    if (config.featureFlags.showAnalyticsPcDropdown) {
-      res.redirect('/analytics/protected-characteristic?characteristic=age')
```
ushkarev, line 250: i would leave a version of this route in case that just redirects to the new one

neilmendum: Good point - addressed

ushkarev (approved):

PR #183, opened 2022-05-20
INC-567: Fix code smells

ushkarev:
…as suggested by SonarCloud.io

### DIFF
--- a/server/services/healthCheck.test.ts
+++ b/server/services/healthCheck.test.ts
@@ -36,22 +36,18 @@ describe('Healthcheck', () => {
 
 function successfulCheck(name: string): HealthCheckService {
   return () =>
-    new Promise((resolve, _reject) => {
-      resolve({
-        name: `${name}`,
-        status: 'ok',
-        message: 'some message',
-      })
+    Promise.resolve({
+      name: `${name}`,
+      status: 'ok',
+      message: 'some message',
     })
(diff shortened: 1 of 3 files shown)


server/services/healthCheck.test.ts
```diff
@@ -36,22 +36,18 @@ describe('Healthcheck', () => {
 
 function successfulCheck(name: string): HealthCheckService {
   return () =>
-    new Promise((resolve, _reject) => {
-      resolve({
-        name: `${name}`,
-        status: 'ok',
-        message: 'some message',
-      })
+    Promise.resolve({
```
xoen, line 39: 👌 

I see now, what I was talking about earlier doesn't apply here because this is an anonymous arrow function so this makes sense (if let's say this was an `async` function, you could return a value directly without having to wrap in `Promise.resolve()`, but this doesn't apply here)

xoen (approved):

server/services/healthCheck.test.ts
ushkarev: true, the whole thing could instead be

```typescript
async function successfulCheck(name: string): Promise<HealthCheckService> {
  return {
    name,
    status: 'ok',
    message: 'some message',
  }
}
```

but i don't want to diverge too much from what was in the template repo.

PR #186, opened 2022-05-20
INC-567: Type-check chart IDs

ushkarev:


server/routes/analyticsChartTypes.ts
```diff
@@ -0,0 +1,29 @@
...
+  'incentive-levels-by-religion',
+  'incentive-levels-by-sexual-orientation',
+  'entries-by-age',
+  'entries-by-disability',
+  'entries-by-ethnicity',
+  'entries-by-religion',
+  'entries-by-sexual-orientation',
+] as const
+
+export type ChartId = typeof chartIds[number]
```
xoen, line 29: I'm sure it's a stupid question but what's with the `typeof chartIds[number]`, particularly the `number` bit?

ushkarev: `chartIds` is a real JS value, not a type. it's an array of strings, but due to the `as const` at the end, the actual type of `chartIds` is:

```typescript
readonly ['entries-by-location', 'prisoners-with-entries-by-location', 'trends-entries', …]
```
which is still a kind of array so can be indexed by a number. `typeof chartIds[number]` says "what type would you get if you indexed this array by a number?". essentially, it's any one of it's members:

```typescript
'entries-by-location' | 'prisoners-with-entries-by-location' | 'trends-entries' | …
```

ushkarev: it mimics `keyof typeof someConcreteValue` to pull out the properties, but for members of an array

PR #190, opened 2022-05-24
INC-572: Added chart for Incentive levels by PC (part of story)

xoen:
This is behind a `showAnalyticsPcTrends` feature flag.

**NOTE**: Currently there is no drop down and the PC group is hardcoded. Also no tests.


server/services/analyticsService.ts
```diff
@@ -480,6 +480,72 @@ export default class AnalyticsService {
...
+          // it's possible for incentive level to be null
+          incentive
+        if (include) {
+          columnSet.add(incentive)
+        }
+        return include
+      }
+    )
+    if (filteredTables.length === 0) {
+      throw new AnalyticsError(AnalyticsErrorType.EmptyTable, 'Filtered trends report for incentive levels has no rows')
```
ushkarev, line 516: i'd include the PC here so that it can be distinguished from the not-filtered-by-PC error message

.env.sample
```diff
@@ -7,6 +7,7 @@ FEEDBACK_URL=https://example.com/generic-feedback
 FEEDBACK_URL_TABLE=https://example.com/table-feedback
 
 FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE=true
+FEATURE_SHOW_ANALYTICS_PC_TRENDS=true
```
ushkarev, line 10: maybe same for integration test env file?

server/services/analyticsService.ts
```diff
@@ -499,11 +499,13 @@ export default class AnalyticsService {
...
+          // it's possible for incentive level to be null
+          incentive &&
+          // it's possible for characteristic group to be null
+          someCharacteristicGroup &&
           // filter only selected prison
           somePrison === prison &&
-          someCharacteristicGroup.trim() === characteristicGroup &&
-          // it's possible for incentive level to be null
-          incentive
+          someCharacteristicGroup.trim() === characteristicGroup
```
ushkarev, line 508: not necessarily for this PR, but perhaps we should trim all the input values earlier? like in the table-stitching step

.env.sample
xoen: I didn't add it yet as it would default to `false` for the time being and we'll change to `true` when we'll have some integration tests 🤔

xoen: But I'll add as false now in the `.feature.env` 👌

server/services/analyticsService.ts
xoen: Just PC or should we spell out the value for PC/PC group as well you think?

e.g.
a. `Filtered PC trends report for incentive levels has no rows`
b. `Filtered PC trends report for incentive levels for ${protectedCharacteristic}/${characteristicGroup} has no rows`

I think option a. may be enough (I'll made this tweak) but if you think we should be even more explicit let me know.

xoen: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/190/commits/3519f1ceb0a3030f552c0f34210cc0320790e782

.env.sample
xoen: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/190/commits/24f8e8b23465cd8e595ca280dc7aa4bb6e067b5d

PR #194, opened 2022-05-25
INC-572: Dropdown for Incentive levels PC trends

xoen:
- added dropdown
- moved dropdown markup in a little macro used in both PC trends charts
- tweaked dropdowns form to avoid reset of other chart selected PC group on select
- corrected hyphen character in charts content `–`

### DIFF
--- a/server/views/pages/analytics/protectedCharacteristicTemplate.njk
+++ b/server/views/pages/analytics/protectedCharacteristicTemplate.njk
@@ -87,10 +87,50 @@
   </div>
 
   {% if featureFlags.showAnalyticsPcTrends %}
+
+    {% macro pcGroupDropdown(
+      chartId,
+      characteristicName,
+      groupSelectLabel,
+      fieldName,
+      options
+    ) %}
+
+      <form id="form-select-{{ fieldName }}" method="get" action="#heading-{{ chartId }}" novalidate>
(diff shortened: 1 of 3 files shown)


server/views/pages/analytics/protectedCharacteristicTemplate.njk
```diff
@@ -87,10 +87,58 @@
   </div>
 
   {% if featureFlags.showAnalyticsPcTrends %}
+
+    {% macro pcGroupDropdown(
```
neilmendum, line 91: We've been calling these `select` rather than `dropdown` so could name this `pcGroupSelect`

neilmendum (approved):

server/views/pages/analytics/protectedCharacteristicTemplate.njk
xoen: FYI: As discussed, I'll rename the other way around: use "dropdown" here but rename related variables.

xoen: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/194/commits/0eac18a81864fb7d98783d9dc6a15501d8b366b7

neilmendum (approved):
LGTM

PR #197, opened 2022-05-25
INC-575: unit tests

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-575

### DIFF
--- a/server/services/analyticsService.test.ts
+++ b/server/services/analyticsService.test.ts
@@ -615,4 +615,53 @@ describe('AnalyticsService', () => {
       expect(columns).toEqual(['Positive', 'Negative', 'Both', 'None'])
     })
   })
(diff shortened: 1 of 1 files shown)


ushkarev (approved):

xoen:
FYI: @ushkarev I've added some tests for the router and most of these things would be tested there: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/199

neilmendum:
> that the 15-17 age group is conditional based on prison (service)

@ushkarev I think this one is covered by @xoen's https://github.com/ministryofjustice/hmpps-incentives-ui/pull/199 - we don't do any filtering in `getBehaviourEntryTrendsByProtectedCharacteristic` of age group

PR #203, opened 2022-05-30
INC-588: Increase k8s resource limit

xoen:
The `LimitRange` k8s resource in the `prod` namespace has been updated
but the limit on the Pods also need to change accordingly.

See: `generic-service` helm chart's values file:
https://github.com/ministryofjustice/hmpps-helm-charts/blob/main/charts/generic-service/values.yaml#L73-L83


neilmendum (approved):
LGTM

PR #215, opened 2022-06-07
INC-600: Prison group/national switch page & tile to get there

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-600

Initially I was going to call the page `regionalNationalSelection` then looking again at the design it talks about _groups_ rather than _regions_

Right now it has _National_ hardcoded - see TODOs in code


server/views/pages/analytics/changePrisonGroup.njk
```diff
@@ -0,0 +1,31 @@
+{% extends "../../partials/analytics/layout.njk" %}
+
+{% from "govuk/components/select/macro.njk" import govukSelect %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
```
ushkarev, line 4: super minor point: `govukBackLink` isn't used

server/views/pages/analytics/changePrisonGroup.njk
```diff
@@ -0,0 +1,31 @@
+{% extends "../../partials/analytics/layout.njk" %}
```
ushkarev, line 1: i think you want `../../partials/layout.njk` cos doesn't the analytics one bring in the levels/entries/PC tabs?

server/routes/analyticsRouter.ts
```diff
@@ -87,11 +87,38 @@ async function transformAnalyticsError<R>(reportPromise: Promise<R>): Promise<R>
...
+        backUrl: '/',
+      })
+    })
+
+    post('/select-prison-group', async (req, res) => {
+      const { prisonGroup } = req.body
+
+      if (!prisonGroup) {
+        logger.error(req.originalUrl, 'prisonGroup is missing')
+        res.redirect('/analytics/select-prison-group')
```
ushkarev, line 113: if you're being super fancy, you could even `res.render('pages/analytics/changePrisonGroup.njk' …` instead of a redirect and put an error message on the select field. probs not worth it *unless* we allow an empty choice in the select element (i.e. a user could actually make a mistake)

server/views/pages/analytics/changePrisonGroup.njk
neilmendum: Good spot. Also removed the import from `changeLocation.njk`

server/views/pages/analytics/changePrisonGroup.njk
neilmendum: Good spot

ushkarev (approved):

PR #216, opened 2022-06-08
INC-601: prison group header and link to switch page

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-601

Thought that we can add tests when we build INC-599

Here's a preview when I simulated passing `prisonGroup = 'National'` to `protectedCharacteristicTemplate`:

group header and link updated

### DIFF
--- a/server/views/partials/analytics/layout.njk
+++ b/server/views/partials/analytics/layout.njk
@@ -2,6 +2,15 @@
 {% from "moj/components/sub-navigation/macro.njk" import mojSubNavigation %}
 
 {% block content %}
+  {% if prisonGroup %}
+    <div class="govuk-!-margin-bottom-4">
+      <h1 class="govuk-heading-xl govuk-!-margin-top-0 govuk-!-margin-bottom-0">
+        {{ prisonGroup }}
+      </h1>
+      <a href="/analytics/select-prison-group" class="govuk-link">Select another prison group</a>
+    </div>
+  {% endif %}
+
   {{ mojSubNavigation({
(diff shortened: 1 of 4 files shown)


server/views/partials/layout.njk
```diff
@@ -41,7 +41,13 @@
 
 {% block header %}
   {% include "./header.njk" %}
-  {% include "./locationBar.njk" %}
+
+  {% if prisonGroup %}
```
neilmendum, line 45: I thought about using `not prisonGroup`, but decided to use the same logic as `server/views/partials/analytics/layout.njk`

ushkarev (approved):

server/views/partials/analytics/layout.njk
```diff
@@ -2,6 +2,15 @@
 {% from "moj/components/sub-navigation/macro.njk" import mojSubNavigation %}
 
 {% block content %}
+  {% if prisonGroup %}
+    <div class="govuk-!-margin-bottom-4">
+      <h1 class="govuk-heading-xl govuk-!-margin-top-0 govuk-!-margin-bottom-0">
```
neilmendum, line 7: Discussed with a11y folks (who overheard @ushkarev and me discussing it) whether multiple `h1` tags on a page is bad practice. It was decided that because this header is above the navigation tabs it was acceptable

Other option was to have this as `h2` which follows [Example 2: Main heading after navigation
](https://www.w3.org/WAI/tutorials/page-structure/headings), but that was potentially confusing to have `h2, h1, h2, etc`

PR #220, opened 2022-06-13
INC-596: Filtering 'analytics' data by PGD region or national

xoen:
Added ability to filter by PGD region or nothing (national) to analytics service methods.

Examples:

```JavaScript
// Filter by prison
analyticsService.getBehaviourEntriesByLocation(Filtering.byPrison('MDI')
// Filter by PGD region
analyticsService.getBehaviourEntriesByLocation(Filtering.byPrison('Yorkshire')
// National
analyticsService.getBehaviourEntriesByLocation(Filtering.national()
```


server/services/analyticsService.ts
```diff
@@ -35,6 +35,35 @@ import {
 } from './analyticsServiceUtils'
 import PrisonRegister from '../data/prisonRegister'
 
+type Query =
+  | { filterColumn: 'prison'; filterValue: string; groupBy: 'wing' }
```
neilmendum, line 39: Minor one - could extract `prison`, `wing`, `pgd_region` to constants as we use them in multiple places in `Query` and `Filtering`

xoen: Hopefully this is similar to what you had in mind? https://github.com/ministryofjustice/hmpps-incentives-ui/pull/220/commits/cb85cc88137839dcba246eceac107c80282c82a1

I thought it would be a bit overkill, but I don't mind it either ☺️

neilmendum (approved):

PR #234, opened 2022-06-21
Filter out rows with blank PGD region.

xoen:
This shouldn't happen and the `null` values were causing an exception
when sorting by this field.

Kinda related to INC-596


xoen:
This shouldn't be required, discussed with Data Scientists and they confirmed it is an issue they resolved in newer source files.

CC: @ushkarev

PR #235, opened 2022-06-22
INC-612: Add file system cache for analytics tables

ushkarev:
…replacing in-memory caching since several gigabytes are now required to store all necessary chart data.
Experimentally enabled only in `preprod`.


server/services/stitchedTablesCache.test.ts
```diff
@@ -1,190 +1,195 @@
...
+    )
   })
+})
 
-  describe('clear()', () => {
-    const columnsToStitch = ['prison']
+describe.each([
+  ['in-memory', MemoryStitchedTablesCache],
+  ['file system', FileStitchedTablesCache],
+])('Stitched table caches', (cacheDescription, CacheConstructor) => {
```
ushkarev, line 23: NB: the exact same test methods are now used for both types of caches; diff is purely whitespace changes

helm_deploy/values-preprod.yaml
```diff
@@ -24,6 +24,7 @@ generic-service:
     FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE: "true"
     FEATURE_SHOW_ANALYTICS_PC_TRENDS: "false"
     FEATURE_SHOW_REGIONAL_NATIONAL: "false"
+    FEATURE_FS_CACHE: "true"
```
ushkarev, line 27: enabling in `preprod` as nobody ever uses it

server/services/stitchedTablesCache.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  set(cacheKey: string, value: CachedValue): Promise<void> {
+    this.cache[cacheKey] = value
+    return Promise.resolve()
+  }
+}
+
+/**
+ * File system cache of the stiched tables values.
+ * Each cached value is stored as a V8-serialised file in a temporary directory.
```
ushkarev, line 117: V8 serialisation is used so that `Date` objects survive the round trip

server/routes/analyticsRouter.ts
```diff
@@ -87,6 +88,8 @@ async function transformAnalyticsError<R>(reportPromise: Promise<R>): Promise<R>
   }
 }
 
+export const cache = new StitchedTablesCache()
```
xoen: That's nicer, there would still be one instance created and shared across all the request handlers per running process AFAICT.

ushkarev: yup, that's exactly the plan: a global cache per-process (there's only 1 process per pod)

xoen (approved):
nice one ❤️

PR #242, opened 2022-06-24
Remove unnecessary `govuk-link` class from links

ushkarev:
… as it adds no styles given `$govuk-global-styles: true`


neilmendum (approved):
LGTM

PR #246, opened 2022-06-27
Use node16 recommended typescript compiler options

ushkarev:
…with the exception of strict mode: it was not enabled previously. The only change is the `target` increasing from `es2018` to `es2021` – which could slightly improve the compiled code as typescript will use fewer workarounds.


tsconfig.json
```diff
@@ -1,18 +1,15 @@
 {
+  "extends": "@tsconfig/node16/tsconfig.json",
   "compileOnSave": true,
   "compilerOptions": {
-    "target": "es2018",
-    "module": "commonjs",
-    "moduleResolution": "node",
+    "strict": false,
```
ushkarev, line 5: i'd prefer strict mode to be enabled. but the template didn't use strict mode so now loads of types throughout the app would be incorrect in strict mode.

neilmendum (approved):
LGTM

Could be worth porting to https://github.com/ministryofjustice/hmpps-template-typescript?

ushkarev:
>Could be worth porting 
i wasn't sure if i wanted to get into a debate over the `strict` flag :)

PR #247, opened 2022-06-28
INC-612: Enable file-based cache in all environments

ushkarev:
tried it out in `preprod` with [oha](https://github.com/hatoo/oha) and it behaved pretty well. admittedly, the data volumes are much smaller. memory remained reasonable whilst CPU spiked.


neilmendum (approved):
LGTM

PR #250, opened 2022-06-29
INC-638: Add further fallback colours to the chart palette

ushkarev:
… which should not be needed as the source data is pre-processed before it reaches this app and should have no incentive levels other than:
• Basic
• Standard
• Enhanced
• Enhanced 2
• Enhanced 3
• and a catch-all-others: “New incentive”
I.e. 6 colours should have been sufficient anyway

### DIFF
--- a/server/utils/analytics.ts
+++ b/server/utils/analytics.ts
@@ -10,6 +10,7 @@ export const palette = [
   'app-chart-colour--yellow',
   'app-chart-colour--pink',
   'app-chart-colour--mid-grey',
+  // colours below are not expected to be used: presence in chart indicates unforeseen values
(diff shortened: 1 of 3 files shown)


server/utils/analytics.ts
```diff
@@ -35,7 +37,7 @@ export function makeChartPalette(columns: string[]): Colour[] {
     if (column === 'None') {
```
neilmendum, line 37: Do we want to explicitly set _Enhanced 3_ to pink?
```
if (column === 'Enhanced 3') {
    return takeColour('app-chart-colour--pink')
}
```

neilmendum (approved):

PR #263, opened 2022-07-14
INC-667: List prisons by region on about page using source data

ushkarev:
…as opposed to hard-coding so that the copy will not need updates as prisons change in future

Note: While it was not caused by these changes, a potential problem with some jest tests was found during this work.

All tests that involve the _mocked_ `ZendeskClient` (inside homeRouter.test.ts and analyticsRouter.test.ts) cause jest to warn about open handles when run in isolation. When the entire suite is run, then it's slow enough for handles to be closed by the time jest is ready to finish. In isolation, jest thinks that open handles may prevent exiting, but it exits immediately. It appears as though there is a lost promise somewhere, perhaps the chain of async request handlers made with `route.all`.

But all tests pass correctly, the site behaves as expected and jest does indeed exit happily.


neilmendum (approved):
LGTM

PR #264, opened 2022-07-14
INC-666: Fix population labels for national/regional trends charts

xoen:
The "total" row incorrectly showed "Prison population" at national and
regional level while it should say "National population" and "Total group
population" respectively.


neilmendum (approved):
LGTM

PR #275, opened 2022-08-12
INC-674: Add hotjar tracking code

xoen:
This is added just to Analytics pages.

**NOTE**: CSP may need some tweaks, [hotjar CSP docs](https://help.hotjar.com/hc/en-us/articles/115011640307-Content-Security-Policies) say to add `'unsafe-inline'` to `script-src`/`style-src` but I want to give it a try without it.


neilmendum (approved):
LGTM

PR #279, opened 2022-08-15
INC-674: Added hotjar CSP rules to default-src

xoen:
as suggested in their CSP docs: https://help.hotjar.com/hc/en-us/articles/115011640307-Content-Security-Policies


server/middleware/setUpWebSecurity.ts
```diff
@@ -18,33 +18,24 @@ export default function setUpWebSecurity(): Router {
     helmet({
       contentSecurityPolicy: {
         directives: {
-          defaultSrc: ["'self'"],
-          formAction: ["'self'", new URL(config.apis.hmppsAuth.url).hostname],
-          scriptSrc: [
+          defaultSrc: [
```
xoen, line 21: diff on GH is a bit confusing, I've removed existing hotjar entries and moved everything in `defaultSrc`, rest should be as it was before trying to add hotjar.

neilmendum (approved):

PR #283, opened 2022-08-18
INC-512: Use new column for wing locations _codes_

ushkarev:
• the `wing` column in the source data is now called `location_code` and there is a new, longer, description column `location_desc` – `location_desc` is not yet propagated into stitched data tables nor displayed
• all sample/test data has been updated despite only 2 tables having this column change – this is why so many tests needed updates

essentially, there is no change as yet to what's displayed to users.

NB: this change is not backwards compatible in that old data files (using `wing`) will now not work. this is ok as there’s currently no way to browse the historical data dumps in the site


neilmendum (approved):
LGTM

PR #284, opened 2022-08-19
Upgrade to node 16.17

ushkarev:
…and maintain formatting of dates to match older versions: v16.17 is built against a new ICU library which formats dates differently (`19 August 2022 at 10:12` instead of `19 August 2022, 10:12`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ parameters:
 
   node-version:
     type: string
-    default: 16.16-browsers
+    default: 16.17-browsers
 
   python-version:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - checkout
       - run:
           name: Update npm
-          command: npm install -g npm@latest
+          command: sudo npm install -g npm@latest
```
ushkarev, line 48: global package location is no longer writable by the default user as of `cimg/node:16.17-browsers`

PR #285, opened 2022-08-19
INC-512: Show longer location names in prison-level analytics views

ushkarev:


xoen (approved):
LGTM

ushkarev:
it's a real shame that we have to squash this as the separate commits make the PR much clearer

PR #302, opened 2022-09-30
Sync `hmpps-template-typescript` and update dependencies

ushkarev:


xoen (approved):

ushkarev:
FYI: app insights still works in `dev` so should be all fine

PR #321, opened 2022-11-04
INC-892: Remove prisoner image ID from contract with incentives-api

ushkarev:
…now that this app loads the photos using prisoner number


Mjwillis (approved):
LGTM

PR #328, opened 2022-11-15
INC-924: Feature flag V2 reviews table by prison

ushkarev:


server/config.ts
```diff
@@ -135,7 +135,7 @@ export default {
     addTestErrorEndpoint: flag('FEATURE_ADD_TEST_ERROR_ENDPOINT', false),
     // Whether to hide the 'Days since last review'/'Days on level' columns
     hideDaysColumnsInIncentivesTable: flag('FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE', false),
-    newReviewsTable: flag('FEATURE_NEW_REVIEWS_TABLE', false),
+    newReviewsTable: get('FEATURE_NEW_REVIEWS_TABLE', '').split(','),
```
xoen, line 138: I would have called _something like_ `prisonsWithNewReviewsTable` but given this is probably not gonna be long lived it doesn't matter that much :)

xoen (approved):
LGTM

PR #330, opened 2022-11-17
INC-843: Update sorting parameters to match incentive-api's contract

ushkarev:
NB: sort & order options have been renamed to use upper snake case because incentives-api follows Java & Spring Data convention

Depends on [api#270](https://github.com/ministryofjustice/hmpps-incentives-api/pull/270)

### DIFF
--- a/server/routes/reviewsTable.ts
+++ b/server/routes/reviewsTable.ts
@@ -114,10 +114,10 @@ function parseSorting(sortString: string | undefined, orderString: string | unde
     // these columns prefer descending order:
     if (['positiveBehaviours', 'negativeBehaviours', 'acctStatus'].includes(sortString)) {
       // eslint-disable-next-line no-param-reassign
-      orderString = 'descending'
+      orderString = 'desc'
     } else {
       // eslint-disable-next-line no-param-reassign
-      orderString = 'ascending'
+      orderString = 'asc'
     }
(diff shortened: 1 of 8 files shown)


server/routes/reviewsTable.ts
```diff
@@ -106,18 +106,18 @@ function parseSorting(sortString: string | undefined, orderString: string | unde
...
     // these columns prefer descending order:
-    if (['positiveBehaviours', 'negativeBehaviours', 'acctStatus'].includes(sortString)) {
+    if (['POSITIVE_BEHAVIOURS', 'NEGATIVE_BEHAVIOURS', 'ACCT_STATUS'].includes(sortString)) {
       // eslint-disable-next-line no-param-reassign
-      orderString = 'descending'
+      orderString = 'DESC'
     } else {
       // eslint-disable-next-line no-param-reassign
-      orderString = 'ascending'
+      orderString = 'ASC'
```
xoen, line 120: I'm guessing the API doesn't care about the case of the `sort` query param?

ushkarev: actually, turns out it's extremely fussy about case. that's why i've renamed all these params. i would prefer lower camel case, but that makes the incentives-api more awkward to build and wouldn't follow Spring Data convention

PR #332, opened 2022-11-22
INC-931: Use 0-based page numbering for incentives reviews api

ushkarev:
In the user interface, pages are still numbered starting from 1 as users would expect. However, incentives-api now starts counting pages from 0.

Depends on [api#277](https://github.com/ministryofjustice/hmpps-incentives-api/pull/277)


neilmendum (approved):
LGTM

PR #354, opened 2022-12-19
INC-953: Enable new reviews table for private beta prison groups 1 & 2

ushkarev:
See also: [digital-prison-services#2312](https://github.com/ministryofjustice/digital-prison-services/pull/2312)
Should be deployed on 5 Jan.


xoen (dismissed):
LGTM

I'm guessing GHI, GMI, SWI, UKI, WDI are part of a separate/later batch?

xoen (approved):

PR #362, opened 2023-01-06
INC-998: National policy banner on 'Reviews' page

xoen:
When the user clicks on the 'Dismiss' link a cookie is set to remember to not show this banner (similar to what Check My Diary does on their service)

See Design ticket: https://dsdmoj.atlassian.net/browse/INC-997

**NOTE**: Current iteration points to the GOV.UK policy document directly but there is a plan to add an intermediate page with an easier to understand summary of the policy key rules. We may want to use a different notification id for that message.


server/routes/all.ts
```diff
@@ -39,6 +40,7 @@ export default function routes(userService: UserService): Router {
   router.use('/analytics/:pgdRegionCode([A-Z0-9]{2,5}|National)?', analyticsRouter(standardRouter(userService)))
   router.use('/prisoner-images/:prisonerNumber.jpeg', prisonerImagesRoutes(imageRouter()))
   router.use('/throw-test-error', throwTestErrorRouter(standardRouter(userService)))
+  router.use('/notification/dismiss', notificationRouter(standardRouter(userService)))
```
ushkarev, line 43: in case we add another notification somewhere, i think it'd be good to make the URL a bit more unique to _this_ notification…?

server/routes/notificationRouter.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import request from 'supertest'
+
+import { appWithAllRoutes } from './testutils/appSetup'
+import UserService from '../services/userService'
+
+jest.mock('../services/userService')
+
+const notificationId = 'test-notification-id'
+const expectedCookieName = `notification-banner-${notificationId}`
+const expectedMaxAgeSecs = 52 * 24 * 3600
```
ushkarev, line 11: why 52 days?

server/routes/notificationRouter.ts
```diff
@@ -0,0 +1,28 @@
+import type { NextFunction, Request, RequestHandler, Response, Router } from 'express'
+
+import asyncMiddleware from '../middleware/asyncMiddleware'
+import NotificationService from '../services/notificationService'
+
+const oneYearMs = 52 * 24 * 3600 * 1000
```
ushkarev, line 6: this is only 52 days

server/routes/notificationRouter.test.ts
xoen: Sorry, got this wrong, was meant to be 52 _weeks_, I'll fix.

server/routes/notificationRouter.ts
xoen: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/362/commits/62c47043f7c1e9951f05992842573f1ba19cac62

server/routes/notificationRouter.test.ts
xoen: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/362/commits/62c47043f7c1e9951f05992842573f1ba19cac62

server/views/pages/reviewsTable.njk
```diff
@@ -107,4 +132,27 @@
...
+        headers: {
+          'X-CSRF-Token': csrf,
+        },
+        data: { id: id },
+      })
+        .done(() => notificationBanner.hide())
+        .fail(() => window.location = '/')
+    }
+
+    $('.notification_dismiss_link').on('click', hideBanner)
```
ushkarev, line 156: i think this would be better to put into /assets/application.js

server/views/pages/reviewsTable.njk
```diff
@@ -107,4 +132,27 @@
...
+      $.ajax({
+        type: 'POST',
+        url: '/notification/dismiss',
+        headers: {
+          'X-CSRF-Token': csrf,
+        },
+        data: { id: id },
+      })
+        .done(() => notificationBanner.hide())
+        .fail(() => window.location = '/')
```
ushkarev, line 153: if the notification fails to dismiss, i figure the user might as well stay where they are?

xoen: I wanted to ask for your opinion about this actually.

We could do nothing (on error notification stays), hide anyway (on reload banner would reappear which could be confusing)

xoen: Cool, I've removed the redirect: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/362/commits/c4d978739026b54da598bb830ed1bf09da5391e6

server/views/pages/reviewsTable.njk
xoen: Yeah that's "static" (without anything interpolated/rendered in it). I've moved to `application.js`: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/362/commits/69a1708aae1e107388a1d930ef1e9b432cf2837c

ushkarev (approved):

PR #365, opened 2023-01-09
INC-998: Incraese notification banner max width

xoen:
This is to reduce the change of text being put on the next line too early looking odd.

### DIFF
--- a/assets/sass/application.scss
+++ b/assets/sass/application.scss
@@ -25,3 +25,7 @@ $govuk-page-width: $moj-page-width;
 @import './global';
 @import './incentive-summary';
 @import './pgd-region';
+
+.reviews-policy-notification-banner.govuk-notification-banner p {
+  max-width: 90%;
+}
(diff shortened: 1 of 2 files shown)


assets/sass/application.scss
```diff
@@ -25,3 +25,7 @@ $govuk-page-width: $moj-page-width;
 @import './global';
 @import './incentive-summary';
 @import './pgd-region';
+
+.govuk-notification-banner p {
```
ushkarev, line 29: i don't think it's a great idea to force _all_ notification banners to have a different shape from the GDS spec. would be better to add a class specific to _this_ banner. something like `.reviews-policy-notification-banner.govuk-notification-banner p`

ushkarev (approved):

assets/sass/application.scss
ushkarev: normally, i'm all for consistency! but i'm hoping that if we use this banner component again elsewhere, we'd use it without modification with respect to what GDS designed

PR #379, opened 2023-01-23
INC-1040: Reviews table: 'Basic' tab selected by default

xoen:
When the user didn't select any incentive level yet page will show the 'Basic' level instead of the prison default level. If 'Basic' is not available at the prison for whatever reason the first level will be shown (this behaviour was there before).

Users want to focus on prisoners on 'Basic' so that they can be reviewed and possibly moved to a better incentive level as soon as possible.


neilmendum (approved):
LGTM

PR #385, opened 2023-02-01
INC-1040: Reviews table should show basic level by default

ushkarev:
Minor tweak to #379
• Find basic level using code rather than human-readable description as it's likely to be more permanent
• Make test cases reflect the fact that basic is now the default tab rather than the default level


xoen (approved):
LGTM 👍

PR #392, opened 2023-02-08
INC-1073: Show number of prisoners and overdue by level

xoen:
- Instead of showing the total number of prisoners overdue a review
  the page now breaks these down by Incentive level
- Show the total number of prisoners at each level in parenthesis
  on the tabs

This is using the updated 'reviews' endpoint: https://github.com/ministryofjustice/hmpps-incentives-api/pull/350

See prototype: https://www.figma.com/file/HBC8couSBqGFO5IjH5yyhk/INC-793---Incentives---%23-on-each-incentive-level-%2B-%23-of-overdue-reviews?node-id=615%3A21063&t=gif2TlUcyyZDUgoc-0

### DIFF
--- a/server/routes/reviewsTable.ts
+++ b/server/routes/reviewsTable.ts
@@ -78,7 +78,8 @@ export default function routes(router: Router): Router {
       order,
     })
 
-    const pageCount = Math.ceil(response.reviewCount / PAGE_SIZE)
+    const { reviewCount } = response.levels.find(level => level.levelCode === selectedLevelCode)
+    const pageCount = Math.ceil(reviewCount / PAGE_SIZE)
     const paginationUrlPrefix = `?level=${selectedLevelCode}&sort=${sort}&order=${order}&`
     const paginationParams = pagination(page, pageCount, paginationUrlPrefix)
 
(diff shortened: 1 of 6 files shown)


server/routes/reviewsTable.ts
```diff
@@ -78,7 +78,8 @@ export default function routes(router: Router): Router {
       order,
     })
 
-    const pageCount = Math.ceil(response.reviewCount / PAGE_SIZE)
+    const { reviewCount } = response.levels.find(level => level.levelCode === selectedLevelCode)
```
ushkarev, line 81: a level can have 0 prisoners/reviews and from what i can read here it'd render ok… right?

server/views/pages/reviewsTable.njk
```diff
@@ -17,14 +17,28 @@
...
 
-  <dl class="app-data-list govuk-!-margin-top-6 govuk-!-margin-bottom-8">
-    <dt>Overdue reviews</dt>
-    <dd>
-      {{- overdueCount -}}
-    </dd>
-  </dl>
+  <div class="govuk-grid-row govuk-!-margin-top-8">
+    <div class="govuk-grid-column-full">
+      <h2 class="govuk-heading-l">Overdue reviews</h2>
```
ushkarev, line 22: the h2 doesn't need to be in a row as it is allowed to go full width, but doesn't really matter

ushkarev (approved):

server/routes/reviewsTable.ts
xoen: Yeah, if there are no prisoners at a given level `reviewCount` would be `0` which means `pageCount` would be `1` (ceil(`0/20`) => ceil(`0`))

PR #419, opened 2023-03-08
INC-1126: Show number of days since the last real review in reviews table and enable sorting

ushkarev:
Depends on [incentives-api#372](https://github.com/ministryofjustice/hmpps-incentives-api/pull/372)


server/views/pages/reviewsTable.njk
```diff
@@ -73,6 +73,14 @@
       {% endif %}
     {% endset %}
 
+    {% set daysSinceLastReview %}
+      {% if review.daysSinceLastReview is null %}
+        Not reviewed
```
ushkarev, line 78: Shown whenever there has not been a "real" in-person review. This is normally shown only for people who are new to prison, but it's possible in other situations (e.g. IEP records were deleted in NOMIS).

xoen (approved):

PR #435, opened 2023-03-27
INC-1164: April 3rd: Release 'Reviews' table to all prisons

xoen:
I've also changed this feature flag in `dev`/`preprod` as there is no point to keep the old table around for some prisons there (and at the moment there is still a way to check the old table if necessary).

**NOTE**: This can be merged on Monday 3rd April.


ushkarev (approved):

neilmendum (approved):
LGTM

PR #437, opened 2023-03-28
INC-1177: Edit incentive levels and their per-prison associated information

ushkarev:
Part 2: write views for QA-ing incentive level apis

Each route is only accessible locally and in `dev` environment and there is no navigable link to these pages.
Modifications require the user to have additional roles.

As with part 1 (#436), there are no unit or integration tests because these are just mockups: proper designs and testing will follow.

### DIFF
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -15,6 +15,7 @@ export interface PrisonIncentiveLevel {
   levelCode: string
   levelDescription: string
   active: boolean
+  defaultOnAdmission: boolean
 
   remandTransferLimitInPence: number
   remandSpendLimitInPence: number
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -25,6 +25,11 @@ export interface PrisonIncentiveLevel {
   privilegedVisitOrders: number
 }
 
+export type PrisonIncentiveLevelUpdate = Omit<
+  Partial<PrisonIncentiveLevel>,
+  'prisonId' | 'levelCode' | 'levelDescription'
+>
+
 interface IncentivesPrisonerSummary {
   prisonerNumber: string
   bookingId: number
@@ -147,6 +152,17 @@ export class IncentivesApi extends RestClient {
   getPrisonIncentiveLevel(prisonId: string, levelCode: string): Promise<PrisonIncentiveLevel> {
     return this.get({
       path: `/incentive/prison-levels/${encodeURIComponent(prisonId)}/level/${encodeURIComponent(levelCode)}`,
+    })
+  }
+
+  updatePrisonIncentiveLevel(
+    prisonId: string,
+    levelCode: string,
+    data: PrisonIncentiveLevelUpdate,
+  ): Promise<PrisonIncentiveLevel> {
+    return this.patch({
+      path: `/incentive/prison-levels/${encodeURIComponent(prisonId)}/level/${encodeURIComponent(levelCode)}`,
+      data,
     })
   }
 
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -153,6 +153,13 @@ export class IncentivesApi extends RestClient {
     })
   }
 
+  setIncentiveLevelOrder(levelCodes: string[]): Promise<IncentiveLevel[]> {
+    return this.patch({
+      path: `/incentive/level-order`,
+      data: levelCodes as unknown as Record<string, unknown>,
+    })
+  }
+
   getPrisonIncentiveLevels(prisonId: string): Promise<PrisonIncentiveLevel[]> {
     return this.get({ path: `/incentive/prison-levels/${encodeURIComponent(prisonId)}` })
   }
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -8,6 +8,8 @@ export interface IncentiveLevel {
   required: boolean
 }
 
+export type IncentiveLevelUpdate = Omit<Partial<IncentiveLevel>, 'code'>
+
 export interface PrisonIncentiveLevel {
   prisonId: string
   levelCode: string
@@ -129,6 +131,13 @@ export class IncentivesApi extends RestClient {
 
   getIncentiveLevel(code: string): Promise<IncentiveLevel> {
     return this.get({ path: `/incentive/levels/${encodeURIComponent(code)}` })
+  }
+
+  updateIncentiveLevel(levelCode: string, data: IncentiveLevelUpdate): Promise<IncentiveLevel> {
+    return this.patch({
+      path: `/incentive/levels/${encodeURIComponent(levelCode)}`,
+      data,
+    })
   }
 
   getPrisonIncentiveLevels(prisonId: string): Promise<PrisonIncentiveLevel[]> {
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -138,6 +138,13 @@ export class IncentivesApi extends RestClient {
     return this.get({ path: `/incentive/levels/${encodeURIComponent(code)}` })
   }
 
+  createIncentiveLevel(data: IncentiveLevel): Promise<IncentiveLevel> {
+    return this.post({
+      path: '/incentive/levels',
+      data: data as unknown as Record<string, unknown>,
+    })
+  }
+
   updateIncentiveLevel(levelCode: string, data: IncentiveLevelUpdate): Promise<IncentiveLevel> {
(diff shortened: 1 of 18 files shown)


server/routes/incentiveLevels.ts
```diff
@@ -1,32 +1,62 @@
...
 import asyncMiddleware from '../middleware/asyncMiddleware'
+import authorisationMiddleware from '../middleware/authorisationMiddleware'
 import { IncentivesApi } from '../data/incentivesApi'
 
+const role = 'ROLE_MAINTAIN_INCENTIVE_LEVELS'
+const requireRole = authorisationMiddleware([role])
+const hasRole = (res: Response) => {
+  const { authorities: roles = [] } = jwtDecode(res.locals.user.token) as { authorities?: string[] }
+  return roles && roles.includes(role)
+}
```
xoen, line 17: I don't know if this is changing a lot in the subsequent commits but I have a couple of small observations:
- `role` is quite generic for a constant
- `role` seems to only be used inside `hasRole`
- you could make that constant more explicit, e.g. call it `ROLE_MAINTAIN_INCENTIVE_LEVELS` (I know that would be long)
- ...or move it inside `hasRole()`
- `hasRole()` in this context literally mean `canManageIncentiveLevels()`
- however I see there is another `hasRole()` function which is doing pretty much the same thing. Maybe this could be extracted in a separate module responsible for role checking?
- _Maybe_ these roles constants could be moved in this module in an enum or something
- I also see a similar logic is in the `authorisationMiddleware`. This comes from the TS template IIRC so I don't know if you want to touch it much but it could also reuse this `hasRole()` function

server/data/incentivesApi.ts
```diff
@@ -25,6 +25,11 @@ export interface PrisonIncentiveLevel {
   privilegedVisitOrders: number
 }
 
+export type PrisonIncentiveLevelUpdate = Omit<
```
xoen, line 29: I do wonder if for these `*Update` types `Pick<Type, Keys>` would be a better alternative to
`Omit<>`.

The idea is that we have a list of fields you can update rather than get all the list and remove some. I know the `PrisonIncentiveLevel` type is probably not going to change much but in theory if a new property is added it would be allowed inside `PrisonIncentiveLevelUpdate`

Although I appreciate this is a bit academic as we the API would respond restrict things (or ignored), etc, so don't worry too much about this. It's just a thought ☺️

server/routes/incentiveLevels.ts
ushkarev: nah that's fair. i just wrote it generically so the code could be copied exactly into the other route, with only the constant value being different.

initially, i had wanted so higher-up middleware to set the roles on the user object (somewhere) but thought that would be changing too much.

will look at better naming

server/data/incentivesApi.ts
ushkarev: hmmmm i thought the opposite weirdly: to omit the read-only / irrelevant fields so that future additions are automatically included in the update.

also: omit 3 fields vs pick 7… more to write in the reversed version!

server/routes/incentiveLevels.ts
```diff
@@ -26,7 +29,7 @@ export default function routes(router: Router): Router {
     return res.render('pages/incentiveLevels.njk', { messages: req.flash(), canEdit, incentiveLevels })
   })
 
-  get('/:levelCode', async (req, res) => {
+  get('/view/:levelCode', async (req, res) => {
```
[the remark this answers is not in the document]
xoen: Oh it's because we're want to distinguish between the page to view a level and the page to _edit_ a level, ignore me 👍

server/views/pages/prisonIncentiveLevelForm.njk
```diff
@@ -0,0 +1,135 @@
...
+        {% call govukFieldset({
+          legend: {
+            text: "Visit orders",
+            classes: "govuk-fieldset__legend--m"
+          }
+        }) %}
+
+          {% set formId = "visitOrders" %}
+          {% set field = form.getField(formId) %}
+          {{ govukInput({
```
xoen, line 92: We could use `inputmode: "numeric"` for the visits, as suggested here: https://design-system.service.gov.uk/components/text-input

ushkarev: good idea

server/routes/prisonIncentiveLevels.ts
```diff
@@ -25,13 +25,19 @@ export default function routes(router: Router): Router {
 
     const { id: prisonId, name: prisonName } = res.locals.user.activeCaseload
     const canEdit = hasRole(res)
-    const prisonIncentiveLevels = await incentivesApi.getPrisonIncentiveLevels(prisonId)
+    const [incentiveLevels, prisonIncentiveLevels] = await Promise.all([
+      incentivesApi.getIncentiveLevels(),
+      incentivesApi.getPrisonIncentiveLevels(prisonId),
+    ])
+    const activeLevelCodes = new Set(prisonIncentiveLevels.map(prisonIncentiveLevel => prisonIncentiveLevel.levelCode))
+    const inactiveIncentiveLevels = incentiveLevels.filter(incentiveLevel => !activeLevelCodes.has(incentiveLevel.code))
```
xoen, line 33: I'm wondering if I'm missing something in this few lines of code.

The `getIncentiveLevels()` method seem to return only the active ones by default, the method call above doesn't pass `withInactive = true` so this seems to be the case.

However here the code seems to imply it's trying to find the levels inactive at the prison by doing the difference between all the levels (including inactive) and the ones active at the prison...but `incentiveLevels` wouldn't include the inactive in this case right?

Sorry if I'm missing something fundamental here.

server/data/incentivesApi.ts
```diff
@@ -153,6 +153,13 @@ export class IncentivesApi extends RestClient {
     })
   }
 
+  setIncentiveLevelOrder(levelCodes: string[]): Promise<IncentiveLevel[]> {
+    return this.patch({
+      path: `/incentive/level-order`,
+      data: levelCodes as unknown as Record<string, unknown>,
```
xoen, line 159: Wasn't this endpoint getting a list of Incentive levels codes, e.g. a `string[]`?

What's with the `Record<string, unknown>`?

server/routes/incentiveLevels.ts
```diff
@@ -61,6 +61,34 @@ export default function routes(router: Router): Router {
   router.get('/activate/:levelCode', requireRole, asyncMiddleware(activateDeactivate(true)))
   router.get('/deactivate/:levelCode', requireRole, asyncMiddleware(activateDeactivate(false)))
 
+  const moveLevel: { (move: 'up' | 'down'): RequestHandler } = move => async (req, res) => {
```
xoen, line 64: May be worth adding a small comment to define "up" and "down" here and avoid any confusion.

As discussed briefly offline "up"/"down" on the list is a way of seeing this. There is also "up" as in "decrease sequence" 🙃. 

Although I appreciate:
- it's relative
- sequence/order is internal now
- UI is a placeholder and things will probably change

I think a comment doesn't hurt here :)

xoen (approved):

server/data/incentivesApi.ts
ushkarev: yeah the api endpoint wants `string[]` but the superclass of this thing does not explicitly allow `any` or `unknown`. what it does do is provide a default value of `{}` which TS decides is `Record<string, unknown>`.

what i don't quite get is why `Omit<Partial<IncentiveLevel>, 'code'>` satisfies `Record<string, unknown>` but `IncentiveLevel` does NOT satisfy `Record<string, unknown>`! they both have properties which are named only by strings and they both have values of some sort

server/routes/incentiveLevels.ts
ushkarev: we don't even know if we will need this route cos it is extremely ui dependant. but _fine_ ;)

server/views/pages/prisonIncentiveLevelForm.njk
```diff
@@ -106,6 +112,8 @@
             id: form.formId + "-" + formId,
```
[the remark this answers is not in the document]
ushkarev: that's a typo :( will fix

server/routes/incentiveLevels.ts
xoen: I'm not suggesting changing anything (as the current concept of up/down is **_not_** wrong).
But I'm thinking maybe add a small comment above here to explain what "up" and "down" _mean_ would avoid confusion :)

xoen (dismissed):

xoen (approved):

PR #440, opened 2023-03-30
INC-1182 Removing old incentives behaviour entries table

ashleygyngell:


ushkarev:
There are a fair few other things can be removed as well. Let's chat later

server/routes/incentivesTable.ts
```diff
@@ -1,41 +0,0 @@
...
-    const agencyId = locationPrefix.split('-')[0]
-
-    const systemToken = await hmppsAuthClient.getSystemClientToken(user.username)
-
-    const behaviorService = new BehaviourService(systemToken)
-    const entries = await behaviorService.getLocationSummary(agencyId, locationPrefix)
-
-    const threeMonthsAgo = daysAgo(90)
-
-    res.render('pages/incentives-table', { entries, threeMonthsAgo, locationPrefix, feedbackUrl })
```
neilmendum, line 30: I think you'll also want to delete `incentives-table.njk`

server/config.ts
```diff
@@ -135,7 +135,6 @@ export default {
     addTestErrorEndpoint: flag('FEATURE_ADD_TEST_ERROR_ENDPOINT', false),
     // Whether to hide the 'Days since last review'/'Days on level' columns
     hideDaysColumnsInIncentivesTable: flag('FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE', false),
```
neilmendum, line 137: I think we can also remove this feature flag, as it was only used by `incentivesTable`

neilmendum (commented):
Looks good! Added two comments

server/config.ts
```diff
@@ -133,9 +133,6 @@ export default {
   },
   featureFlags: {
     addTestErrorEndpoint: flag('FEATURE_ADD_TEST_ERROR_ENDPOINT', false),
-    // Whether to hide the 'Days since last review'/'Days on level' columns
-    hideDaysColumnsInIncentivesTable: flag('FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE', false),
```
ushkarev, line 137: `FEATURE_HIDE_DAYS_COLUMNS_IN_INCENTIVES_TABLE` was only used in the old table, so i think it's correct to remove it.

it still appears in the various helm_deploy/values files and some .env files too so i'd get delete it there too

xoen (approved):
Nice one @ashleygyngell - love those cleanup PRs 😍

PR #441, opened 2023-04-04
Removing unnecessary build (hmpps-template-typescript#172)

ushkarev:
All 3 processes: tsc, sass and copy-views are run by concurrently at start up anyway


xoen (approved):
Nice 👍

PR #443, opened 2023-04-04
INC-1182: Delete references to removed incentives-api endpoint

ushkarev:
Continues on from #440

`incentives-api` has removed the `/incentives-summary/prison/${prisonId}/location/${locationId}` endpoint in [#394](https://github.com/ministryofjustice/hmpps-incentives-api/pull/394) so the UI app should not contain references to it.


server/testData/incentivesApi.ts
```diff
@@ -1,81 +1,6 @@
...
-            provenAdjudications: 0,
-          },
-        ],
-      },
-    ],
-  }
-}
-
-export function getTestIncentivesReviews(): IncentivesReviewsResponse {
+export default function getTestIncentivesReviews(): IncentivesReviewsResponse {
```
ushkarev, line 3: `eslint` doesn’t let this remain a normal exported function :(

xoen (approved):
Nice 👍

PR #444, opened 2023-04-06
INC-1185: Add admin section to manage incentive level details in a prison

ushkarev:
Users can
• only access this section with an appropriate role (only in `dev` for now)
• see the levels their _current_ prison has active and their associated information
• add more levels if they are globally available and not yet active
• remove a level as long as it is not globally required
• edit associated information for their active levels


Mjwillis (approved):
LGTM

server/routes/prisonIncentiveLevels.test.ts
```diff
@@ -229,6 +231,19 @@ describe('Prison incentive level management', () => {
         .expect(404)
     })
 
+    it('should 404 if level is required globally (POST)', () => {
```
xoen, line 238: Not too bothered about this on the UI side but I wonder if a `400 Bad Request` would be more appropriate in this case?

The level (Standard) is there but it can't be deactivated because it's required globally.

Just a thought :)

server/routes/prisonIncentiveLevels.ts
```diff
@@ -93,11 +93,23 @@ export default function routes(router: Router): Router {
...
+        const [incentiveLevel, prisonIncentiveLevel] = await Promise.all([
+          incentivesApi.getIncentiveLevel(levelCode),
+          incentivesApi.getPrisonIncentiveLevel(prisonId, levelCode),
+        ])
+        if (!prisonIncentiveLevel.active) {
+          throw NotFound('Prison incentive level is already inactive')
+        }
+        if (incentiveLevel.required) {
+          throw NotFound('Incentive level is required globally')
+        }
```
xoen, line 107: See comment above related to this. A `400` may be more appropriate.

xoen (approved):
Nice work 🌟 

I've added a comment about `404` vs `400` when trying to remove a required Incentive level, don't know whether you want to tweak that or you think it's the correct behaviour, it's a minor thing anyway.

server/routes/prisonIncentiveLevels.test.ts
ushkarev: yeah ok. particularly given there's already another error condition that returns a 400.

i'm a little wary of the more esoteric status codes for content returned to a _browser_… i have a vague recollection that they used to show their own error pages instead. anyway, this is a rare edge case that should never be seen by users.

server/routes/prisonIncentiveLevels.ts
ushkarev: done! 26fc4ada68ad421a8b5e25a3587a23a0ba683ce5

server/routes/prisonIncentiveLevels.test.ts
ushkarev: done! 26fc4ada68ad421a8b5e25a3587a23a0ba683ce5

xoen (approved):

PR #450, opened 2023-04-13
Sync upstream template: Cache static resources for 1 hour (hmpps-template-typescript#178)

ushkarev:


ushkarev:
having only 20s cache for static resources is a bit silly

neilmendum (approved):
LGTM

PR #463, opened 2023-04-21
INC-1212: Handle api errors properly when incentive levels are being modified

ushkarev:
…and show a specific custom message when a level cannot be deactivated  because there are prisoners on it.

Turns out that api errors were incorrectly handled before: the error thrown was not directly what the api returned but wrapped inside `SanitisedError`

### DIFF
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -1,5 +1,44 @@
+// eslint-disable-next-line max-classes-per-file
 import config from '../config'
 import RestClient from './restClient'
+
+/**
+ * Structure representing an error response from the incentives api
+ */
+export class ErrorResponse {
+  status: number
+
+  errorCode?: ErrorCode
+
+  userMessage?: string
+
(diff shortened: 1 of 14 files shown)


server/data/incentivesApi.ts
```diff
@@ -1,6 +1,37 @@
...
+
+  moreInfo?: string
+
+  static isErrorResponse(obj: object): obj is ErrorResponse {
+    // TODO: would be nice to make userMessage & developerMessage non-nullable in the api
+    return obj && 'status' in obj && typeof obj.status === 'number'
+  }
+}
+
+export enum ErrorCode {
```
xoen, line 22: Do you reckon is worth adding a note referring to the enum in the API where these are defined?

xoen (approved):

server/data/incentivesApi.ts
ushkarev: done

xoen (approved):

PR #466, opened 2023-04-21
INC-1232: Add user info to AppInsights request

xoen:
Copied/adapted from this Use of force PR:
https://github.com/ministryofjustice/use-of-force/pull/599

### DIFF
--- a/server/utils/azureAppInsights.ts
+++ b/server/utils/azureAppInsights.ts
@@ -1,5 +1,11 @@
-import { setup, defaultClient, TelemetryClient, DistributedTracingModes } from 'applicationinsights'
+import { Contracts, setup, defaultClient, TelemetryClient, DistributedTracingModes } from 'applicationinsights'
+import { EnvelopeTelemetry } from 'applicationinsights/out/Declarations/Contracts'
 import applicationVersion from '../applicationVersion'
+
+export type ContextObject = {
+  /* eslint-disable  @typescript-eslint/no-explicit-any */
+  [name: string]: any
+}
 
 function defaultName(): string {
   const {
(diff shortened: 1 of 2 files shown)


server/utils/azureAppinsights.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      expect(envelope.data.baseData.properties).toStrictEqual({ other: 'things' })
+    })
+
+    it('returns true when not RequestData type', () => {
+      const envelope = createEnvelope({}, 'NOT_REQUEST_DATA')
+
+      const response = addUserDataToRequests(envelope, context)
+
+      expect(response).toStrictEqual(true)
```
ushkarev, line 59: should this be in every test maybe?

server/utils/azureAppInsights.ts
```diff
@@ -1,6 +1,12 @@
-import { setup, defaultClient, TelemetryClient, DistributedTracingModes } from 'applicationinsights'
+import { Contracts, setup, defaultClient, TelemetryClient, DistributedTracingModes } from 'applicationinsights'
+import { EnvelopeTelemetry } from 'applicationinsights/out/Declarations/Contracts'
 import applicationVersion from '../applicationVersion'
 
+export type ContextObject = {
+  /* eslint-disable  @typescript-eslint/no-explicit-any */
```
ushkarev, line 6: i think you should use this form instead

```javascript
// eslint-disable-next-line @typescript-eslint/no-explicit-any
```

because the one you used likely disables the lint for the entire file

server/utils/azureAppinsights.test.ts
xoen: I just copied from the other PR and had a look at the tests, as you said I don't think it really matters :)

server/utils/azureAppInsights.ts
xoen: Oh, missed that, I'll change 👍

xoen: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/466/commits/c7e0d46a7f9e0d541b335b197004dde4db247cfc

ushkarev (approved):

PR #473, opened 2023-04-24
INC-1230: Indicate that visit orders are allowances and not hard limits

ushkarev:
~Vist limits~ → Visit allowance


ashleygyngell (approved):
Nice spot!

PR #475, opened 2023-04-25
INC-1238: Show location-based tiles on home page only when appropriate

ushkarev:
There exist users with case loads that aren’t real locations so it does not make sense to show them sections of the site that only work in prisons.

Such users are:
• LSAs when using their special account (with `CADM_I` role)
• central admins who are never in a prison

The former group can now only see the prison-group analytics charts.
The latter group can now only see the prison-group analytics charts and global incentive level admin section.

---

Effectively:
- all users can see the prison-group analytics charts and about page for visualisations
  - central admins, non-prison-based users, with `MAINTAIN_INCENTIVE_LEVELS` role also see global incentive level admin section
  - all prison-based users also see incentive reviews sections and all analytics charts
    - LSAs, using their normal prison-based account, with `MAINTAIN_PRISON_IEP_LEVELS` role also see prison incentive level admin section

---
Relates to: [DPS#2431](https://github.com/ministryofjustice/digital-prison-services/pull/2431)

### DIFF
--- a/server/routes/home.ts
+++ b/server/routes/home.ts
@@ -5,6 +5,7 @@ import jwtDecode from 'jwt-decode'
 import config from '../config'
 import logger from '../../logger'
 import asyncMiddleware from '../middleware/asyncMiddleware'
+import { PrisonApi } from '../data/prisonApi'
 import ZendeskClient, { CreateTicketRequest } from '../data/zendeskClient'
 import S3Client from '../data/s3Client'
 import AnalyticsService from '../services/analyticsService'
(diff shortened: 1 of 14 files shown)


server/routes/home.ts
```diff
@@ -29,17 +30,28 @@ const getUserRoles = (res: Response): string[] => {
 export default function routes(router: Router): Router {
   const get = (path: string | string[], handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
 
-  get('/', (req, res) => {
+  get('/', async (req, res) => {
     res.locals.breadcrumbs.lastItem.href = undefined
 
-    const userRoles = getUserRoles(res)
+    // a prison case load would have locations, e.g. wings or house blocks
```
Mjwillis, line 36: This will work but its kind of a side effect in that locations are not returned from the `me` endpoint for CADM_I - I thought we would have the caseload ID in the session so could just check that.  However happy to leave as is.

Mjwillis (approved):
See comment about API call to locations,  by happy anyway

server/routes/home.ts
ushkarev: I found that `CADM_I` was the active case load ID (not a role / or maybe _also_ a role) so wasn't sure how else we'd know that the "current session" isn't a proper location…

I got the idea of using these locations from the DPS home page’s tiles. Even before me change, there had been a condition on the locations within the active case load:

https://github.com/ministryofjustice/digital-prison-services/commit/49d98598e22c50d42180818a12f85c14c280c918#diff-f8a9492c5a04c0a52f307b83958be45fd23f684f6dafeb35e41a3df8aeaaa675L78

PR #479, opened 2023-04-26
INC-1243: Enhance admin section to manage incentive levels globally

ushkarev:
Improves on #436 and #437. Users…
• can only access this section with an appropriate role (only in `dev` for now)
• can create new incentive levels at the end of the existing list
• can change the status of existing incentive levels unless they are required in every prison
• _cannot_ update levels’ names (must be done using incentives api)
• _cannot_ set whether any level is required in all prisons (must be done using incentives api)
• _cannot_ reorder levels (must be done using incentives api)

### DIFF
--- a/server/routes/forms/incentiveLevelCreateForm.ts
+++ b/server/routes/forms/incentiveLevelCreateForm.ts
@@ -10,8 +10,12 @@ export default class IncentiveLevelCreateForm extends Form<IncentiveLevelCreateD
     if (!this.data.name || this.data.name.length < 1) {
       this.addError('name', 'The level’s name is required')
     }
-    if (!this.data.code || this.data.code.length < 1) {
-      this.addError('name', 'The level’s code is required')
+    // TODO: add max length limit to `name` & error message once determined
+    //       NOMIS has hard limit of 40, incentives DB is 30
+
+    const codeRE = /^[A-Za-z0-9]{3}$/
+    if (!codeRE.test(this.data.code)) {
+      this.addError('code', 'The level’s code must be 3 letters or numbers')
     }
   }
 }
--- a/server/routes/forms/incentiveLevelCreateForm.ts
+++ b/server/routes/forms/incentiveLevelCreateForm.ts
@@ -1,0 +1,17 @@
+import Form, { type BaseFormData } from './forms'
+
+export interface IncentiveLevelCreateData extends BaseFormData {
+  name: string
+  code: string
+}
+
+export default class IncentiveLevelCreateForm extends Form<IncentiveLevelCreateData> {
+  protected validate(): void {
+    if (!this.data.name || this.data.name.length < 1) {
+      this.addError('name', 'The level’s name is required')
+    }
+    if (!this.data.code || this.data.code.length < 1) {
+      this.addError('name', 'The level’s code is required')
+    }
+  }
+}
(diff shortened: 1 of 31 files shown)


server/routes/forms/incentiveLevelCreateForm.ts
```diff
@@ -0,0 +1,17 @@
...
+  code: string
+}
+
+export default class IncentiveLevelCreateForm extends Form<IncentiveLevelCreateData> {
+  protected validate(): void {
+    if (!this.data.name || this.data.name.length < 1) {
+      this.addError('name', 'The level’s name is required')
+    }
+    if (!this.data.code || this.data.code.length < 1) {
+      this.addError('name', 'The level’s code is required')
```
xoen: This should add the error to `code`, not `name` here right?

server/routes/incentiveLevels.ts
```diff
@@ -284,3 +295,29 @@ export default function routes(router: Router): Router {
...
+async function errorMessageWhenCannotDeactivate(prisonApi: PrisonApi, errorResponse: ErrorResponse): Promise<string> {
+  let message = 'This level cannot be made inactive because some establishments are using it.'
+
+  const prisonNames: string[] = []
+  const prisonIds: string[] | undefined = (errorResponse?.moreInfo ?? '')
+    .split(',')
+    .map(prisonId => prisonId.trim())
+    .filter(prisonName => prisonName)
+  // eslint-disable-next-line no-restricted-syntax
+  for (const prisonId of prisonIds) {
```
xoen, line 411: Probably not much of a difference in practice as I'd hope they'd only try to remove a level only when most of the prisons already stopped using it...

But we could potentially make all these requests at once concurrently as retrieving the information for each of the prisons is independent from the others.

ushkarev: hmm not a bad idea!

server/routes/forms/incentiveLevelCreateForm.ts
ushkarev: fixed in a later commit!
https://github.com/ministryofjustice/hmpps-incentives-ui/pull/479/commits/f83f57118db79dd0efd72b6ac530100a5160cad9#diff-5e5411b99aae4cbc9cee6c93e72f93c8feb62d297bd4a5251f4aeb2a3abf3a50R18

server/routes/incentiveLevels.ts
```diff
@@ -17,6 +17,9 @@ export const manageIncentiveLevelsRole = 'ROLE_MAINTAIN_INCENTIVE_LEVELS'
 export default function routes(router: Router): Router {
   const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
 
+  /*
```
xoen, line 20: Does this need to be `/**` to make JSDoc happy? Not that it matters too much tbh

ushkarev: there's not always a sensible variable or function to attach it to, so i just used multi-line comments.

my plan was to actually move all this junk into several separate functions and then do better jsdoc-style comments. didn't wanna do in this commit because the diff will be crazy.

server/routes/forms/incentiveLevelCreateForm.ts
xoen: oh sorry, I was having a look commit-by-commit 👍

xoen (approved):

server/routes/incentiveLevels.ts
ushkarev: how about this? 0c2a5c2e56fef6c7738e61870b99afec7f9c0419

xoen (approved):

PR #484, opened 2023-04-27
INC-1250: Automatically add any missing required incentive levels to a prison

ushkarev:
This is useful for new prisons so that they adopt the standard set of levels.
Additionally, if a prison has no default-for-admissions level (a data inconsistency problem) it will automatically gain Standard.

### DIFF
--- a/server/routes/prisonIncentiveLevels.ts
+++ b/server/routes/prisonIncentiveLevels.ts
@@ -51,6 +51,10 @@ export default function routes(router: Router): Router {
       } catch (error) {
         logger.error(`Missing required incentive levels could not be added to ${prisonName}!`, error)
       }
+    } else {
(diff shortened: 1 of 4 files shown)


server/routes/prisonIncentiveLevels.ts
```diff
@@ -31,22 +34,36 @@ export default function routes(router: Router): Router {
       incentivesApi.getPrisonIncentiveLevels(prisonId),
     ])
 
-    const activeLevelCodes = new Set(prisonIncentiveLevels.map(prisonIncentiveLevel => prisonIncentiveLevel.levelCode))
```
ushkarev, line 34: i would’ve preferred to use sets but they’re pretty pointless if you need to turn them back into arrays to get the difference between two sets

PR #500, opened 2023-05-16
Fix version not appearing in application insights

ushkarev:
Simplified version of [hmpps-template-typescript#190](https://github.com/ministryofjustice/hmpps-template-typescript/pull/190) since application information is constant and might as well all be situated in config.


server/config.ts
```diff
@@ -34,7 +37,23 @@ export interface ApiConfig {
...
 }
 
+export interface ApplicationInfo {
+  applicationName: string
+  buildNumber: string
+  gitRef: string
+}
+
+const packageJson = path.join(__dirname, '../package.json')
+const { name: applicationName } = JSON.parse(fs.readFileSync(packageJson, { encoding: 'utf8' }))
```
ushkarev, line 47: `applicationName` could even have been hard-coded right here…

ushkarev: in fact, i've hard-coded it now so that the built version doesn't need to look in a different location for package.json file

PR #509, opened 2023-05-19
INC-1233: Preliminary incentive level name validation

ushkarev:
Whilst the exact error messages and level name length limits are being determined, we can already put in the hard 30-character limit. That’s the maximum that the incentives-api will allow. As it happens, NOMIS permits up to 40 characters, but this has been deemed excessive.

### DIFF
--- a/server/routes/incentiveLevels.ts
+++ b/server/routes/incentiveLevels.ts
@@ -12,10 +12,6 @@ import IncentiveLevelEditForm from './forms/incentiveLevelEditForm'
 import IncentiveLevelReorderForm from './forms/incentiveLevelReorderForm'
 import IncentiveLevelStatusForm from './forms/incentiveLevelStatusForm'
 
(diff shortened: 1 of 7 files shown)


server/routes/incentiveLevels.ts
```diff
@@ -12,10 +12,6 @@ import IncentiveLevelEditForm from './forms/incentiveLevelEditForm'
 import IncentiveLevelReorderForm from './forms/incentiveLevelReorderForm'
 import IncentiveLevelStatusForm from './forms/incentiveLevelStatusForm'
 
-export function sanitiseIncentiveName(name: string) {
-  return name.trim()
-}
-
```
ushkarev, line 18: I've moved the sanitisation responsibility into the relevant forms. It seems sensible to encapsulate the forms actions within the relevant class rather than rely on the careful use of the form in each situation.

xoen (approved):

PR #513, opened 2023-05-24
INC-1251: Switch over to new api

ushkarev:
…when getting a prison’s set of incentive levels for the reviews table


neilmendum (dismissed):

neilmendum (approved):

ushkarev:
PS: i’ve deliberately left the deprecated functions and stubs in place for other teams to see the upgrade path

PR #519, opened 2023-05-30
INC-1269: Activate the previously-default-for-admissions level

ushkarev:
…when there is no default-for-admissions incentive level in a prison but an inactive one exists. This is to allow the incentives-api to uphold it’s business rules.

Relates to [api#456](https://github.com/ministryofjustice/hmpps-incentives-api/pull/456)


server/routes/prisonIncentiveLevels.test.ts
```diff
@@ -274,6 +274,45 @@ describe('Prison incentive level management', () => {
...
+          }),
+      )
+
+      return request(app)
+        .get('/prison-incentive-levels')
+        .set('authorization', `bearer ${tokenWithNecessaryRole}`)
+        .expect(res => {
+          expect(res.redirect).toBeTruthy()
+          expect(incentivesApi.updatePrisonIncentiveLevel).toHaveBeenCalledTimes(4)
+          expect(incentivesApi.updatePrisonIncentiveLevel).toHaveBeenNthCalledWith(1, 'MDI', 'ENH', {
```
xoen, line 306: starting with `1`, weird...

server/routes/prisonIncentiveLevels.ts
```diff
@@ -483,13 +483,19 @@ async function ensureDefaultLevelExists(
...
   prisonId: string,
   prisonIncentiveLevels: PrisonIncentiveLevel[],
-  defaultLevelCode = 'STD',
 ) {
   if (!prisonIncentiveLevels.some(prisonIncentiveLevel => prisonIncentiveLevel.defaultOnAdmission)) {
+    logger.warn(`${prisonId} is missing a default level`)
+    const defaultLevelCode =
+      (await incentivesApi.getPrisonIncentiveLevels(prisonId, true))
+        .filter(prisonIncentiveLevel => prisonIncentiveLevel.defaultOnAdmission)
+        .map(prisonIncentiveLevel => prisonIncentiveLevel.levelCode)?.[0] ?? 'STD'
```
xoen, line 492: What do you think about using `.find()` instead?

I haven't tested this so it may be wrong but you should be able to write something along the lines of

```
.find(prisonIncentiveLevel => prisonIncentiveLevel.defaultOnAdmission)?.levelCode ?? 'STD'
```

which I think reads a bit better, especially as you wouldn't have to use the `map` and the `.[0]` indexing.

server/routes/prisonIncentiveLevels.ts
```diff
@@ -483,13 +483,19 @@ async function ensureDefaultLevelExists(
   incentivesApi: IncentivesApi,
   prisonId: string,
   prisonIncentiveLevels: PrisonIncentiveLevel[],
-  defaultLevelCode = 'STD',
 ) {
   if (!prisonIncentiveLevels.some(prisonIncentiveLevel => prisonIncentiveLevel.defaultOnAdmission)) {
+    logger.warn(`${prisonId} is missing a default level`)
+    const defaultLevelCode =
+      (await incentivesApi.getPrisonIncentiveLevels(prisonId, true))
```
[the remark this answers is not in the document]
ushkarev: aah if only this were python or kotlin :)

xoen (approved):
Left a couple of minor comments on style (that you may or may not like and possibly ignore if you want) but this looks good to me :)

xoen (dismissed):
Left a couple of minor comments on style (that you may or may not like and possibly ignore if you want) but this looks good to me :)

server/routes/prisonIncentiveLevels.ts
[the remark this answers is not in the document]
xoen: Oh you can't in TS? Between these languages I'm getting so confused 🤣

server/routes/prisonIncentiveLevels.ts
ushkarev: oops i forgot that was a method!

xoen: I like it when the intend is to find one thing (like in this case where you're looking for the "default on admission level") - and it tends to make code nicer in these occurences :)

xoen (approved):

server/routes/prisonIncentiveLevels.ts
ushkarev: totes agree

PR #532, opened 2023-06-14
Point out that `production` config & template parameter is true in dev, preprod _and_ prod

ushkarev:
… and ensure that templates have access to environment & production variables

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -48,7 +48,7 @@ const applicationInfo: ApplicationInfo = {
 
 export default {
   applicationInfo,
-  production,
+  production, // NB: this is true in _all_ deployed environments
   environment: process.env.ENVIRONMENT || 'local',
   https: production,
   staticResourceCacheDuration: '1h',
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -9,24 +9,27 @@ import { calculateTrendsRange, makeChartPalette } from './analytics'
 import format from './format'
 import { daysSince, initialiseName } from './utils'
 
-const production = process.env.NODE_ENV === 'production'
-
 export default function nunjucksSetup(app: express.Express): void {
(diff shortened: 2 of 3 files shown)


server/config.ts
```diff
@@ -48,7 +48,7 @@ const applicationInfo: ApplicationInfo = {
 
 export default {
   applicationInfo,
-  production,
```
xoen, line 51: This is very confusing/misleading, is this needed anywhere?

If it's coming from the TS template maybe we can't remove/rename for some reason.

xoen: Oh at the very least by `nunjucksSetup.ts` 👌

server/utils/nunjucksSetup.ts
```diff
@@ -9,24 +9,27 @@ import { calculateTrendsRange, makeChartPalette } from './analytics'
...
 
   app.locals.asset_path = '/assets/'
   app.locals.applicationName = 'Manage incentives'
-  app.locals.dpsHome = config.dpsUrl
-  app.locals.supportUrl = config.supportUrl
-  app.locals.feedbackUrl = config.feedbackUrl
-  app.locals.googleAnalyticsGa4Id = config.googleAnalytics.ga4MeasurementId
-  app.locals.hotjarSiteId = config.hotjar.siteId
+  app.locals.production = config.production
+  app.locals.environment = config.environment
```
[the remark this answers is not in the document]
xoen: Maybe `environment` was set directly by the routes from the config?

I `locals.production` is not used I'm tempted to say don't set it as it's confusing and someone may use that as an indication this is or isn't `prod`, while `environment` is there for this.

xoen (approved):
So confused by `config.production` being always `true`.

I would consider _**not**_ setting `app.locals.production` and don't expose this to the templates (or at least not implicitly) as it's misleading but I don't mind either way.

PR #536, opened 2023-06-16
Simplify docker-compose settings

ushkarev:
Running the app locally against services in `dev` environment does not require a bespoke docker-compose file. The one for integration testing can be reused so there are fewer settings to manage and keep in sync.


docker-compose-test.yml
```diff
@@ -61,11 +63,11 @@ services:
...
-      - hmpps_int
+      - hmpps
     container_name: inc-ui-wiremock
     restart: always
     ports:
       - '9091:8080'
 
 networks:
-  hmpps_int:
+  hmpps:
```
ushkarev, line 73: there's no need to use a bespoke network for testing; the two docker-compose suites cannot be run concurrently anyway

ashleygyngell (approved):

PR #544, opened 2023-06-26
NN-1284 Locations with same beginning letter (e.g. C or CLIVE) return too many results

Mjwillis:


ushkarev (dismissed):
hmm interesting that the [prison-api swagger docs](https://api.prison.service.justice.gov.uk/swagger-ui/index.html#/users/getMyLocations) don't mention this `subLocations` param

ushkarev:
oh i see, this is dependent on [prison-api#1664](https://github.com/ministryofjustice/prison-api/pull/1664)

server/routes/selectLocation.ts
```diff
@@ -37,13 +37,15 @@ export default function routes(router: Router): Router {
     const allLocations: Array<Location> = await prisonApi.getUserLocations()
     const selectedLocation = allLocations.find(location => location.locationPrefix === locationPrefix)
 
+    const subLocationChar = selectedLocation.subLocations ? '-' : ''
+
     if (!selectedLocation) {
```
ushkarev, line 42: looks like `selectedLocation` can be absent so i'd move the `subLocationChar` line to be below this if block

ushkarev (approved):

PR #580, opened 2023-08-09
INC-1295: Implementing DPS recommendations renaming service from `manage incentives` to `incentives`

ashleygyngell:


server/routes/homeRouter.test.ts
```diff
@@ -147,7 +147,7 @@ describe('Home page', () => {
         })
     })
 
-    it('shows tile to manage incentive levels if user has appropriate role even without having any locations in active case load', () => {
+    it('shows tile to incentive levels if user has appropriate role even without having any locations in active case load', () => {
```
ushkarev, line 150: i think these 2 tests should still say "manage" cos it wasn't the name of the service and now it doesn't read as a sentence properly. they're for checking that the tiles for the admin sections appear

ashleygyngell: You're totally right :)

ushkarev (approved):

PR #606, opened 2023-09-22
INC-1317: Header & Footer update

ashleygyngell:


integration_tests/e2e/login.cy.ts
```diff
@@ -35,24 +35,14 @@ context('SignIn', () => {
...
-    cy.get('.govuk-footer__inline-list .govuk-footer__link:visible').then(footerLinks => {
-      expect(footerLinks.length).equal(2)
-
-      const TsAndCsLink = footerLinks.last()
-      expect(TsAndCsLink).to.contain('Terms and conditions')
+    it('Fallback footer exists with no content', () => {
+      cy.signIn()
+      cy.task('stubDpsComponentsFail')
+      const indexPage = Page.verifyOnPage(HomePage)
+      indexPage.fallbackFooter.should('not.include.text', 'Feedback')
```
ushkarev, line 45: the old footer and the current fallback footer don't contain 'Feedback' so what i think we can do here is `indexPage.fallbackFooter.should('include.text', 'Terms and conditions')` if you do put the `{{ govukFooter(…) }}` into footer.njk

server/data/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import TokenStore from './tokenStore'
+import DpsFeComponentsClient from './dpsComponentsClient'
+
+export const dataAccess = () => ({
+  applicationInfo,
+  hmppsAuthClient: new HmppsAuthClient(new TokenStore(createRedisClient('dpsFrontendComponent'))),
+  componentApiClientBuilder: new DpsFeComponentsClient(),
+})
+
+export type DataAccess = ReturnType<typeof dataAccess>
```
ushkarev, line 23: i'm not sure this is right. we already initialise the App Insights stuff elsewhere. incentives was created from an older template which did not have this file and created services quite differently. i don't know what happens when you try to setup App Insights twice – in fact you seem to have something commented out. i would 1) remove this entire file 2)  remove the `services/index.ts` file too because the hmpps auth client and user service are already made elsewhere 3) just initialise the class in `server/middleware/dpsFrontendComponents.ts`

server/middleware/setUpWebSecurity.ts
```diff
@@ -15,27 +15,34 @@ export default function setUpWebSecurity(): Router {
...
+
+  const scriptSrc = [
+    "'self'",
+    '*.google-analytics.com',
+    '*.googletagmanager.com',
+    (_req: Request, res: Response) => `'nonce-${res.locals.cspNonce}'`,
+  ]
+  const styleSrc = ["'self'", (_req: Request, res: Response) => `'nonce-${res.locals.cspNonce}'`]
+  const imgSrc = ["'self'", 'data:', '*.google-analytics.com', '*.googletagmanager.com']
+  const fontSrc = ["'self'"]
```
ushkarev, line 27: you've removed some directives that are needed from each of these, the hot jar ones and unsafe inlines

PR #616, opened 2023-10-03
Improve REST client by allowing query parameters to be passed for all methods

ushkarev:
… and add PUT & DELETE methods

### DIFF
--- a/server/data/restClient.ts
+++ b/server/data/restClient.ts
@@ -1,7 +1,7 @@
 import { Readable } from 'stream'
 
-import superagent from 'superagent'
 import Agent, { HttpsAgent } from 'agentkeepalive'
+import superagent from 'superagent'
 
 import logger from '../../logger'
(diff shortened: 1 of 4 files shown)


server/data/restClient.ts
```diff
@@ -9,21 +9,20 @@ import type { ApiConfig } from '../config'
...
-interface GetRequest {
+interface BaseRequest {
   path?: string
   query?: object | string
   headers?: Record<string, string>
   responseType?: string
   raw?: boolean
+}
+
+interface Request extends BaseRequest {
```
xoen, line 20: Again, I'm probably missing something but could this `handle404` property be in the `BaseRequest` class and simplify the class hierarchy?

xoen (dismissed):
LGTM 👍 

Do we need these additionally put/delete method tho if they're not used? 🤔

xoen (approved):

PR #626, opened 2023-10-11
INC-1296: Removed vestigial referenced to deprecated endpoint

xoen:
`GET /iep/levels/{prisonId}` has been deprecated for a long time and will be removed at some point soon.

The UI doesn't use this endpoint anymore but there were still a couple of methods related to this endpoint that can be removed.

### DIFF
--- a/server/data/incentivesApi.ts
+++ b/server/data/incentivesApi.ts
@@ -207,13 +207,6 @@ export class IncentivesApi extends RestClient {
     })
   }
 
-  /**
-   * @deprecated use getPrisonIncentiveLevels
-   */
-  getAvailableLevels(agencyId: string): Promise<Level[]> {
-    return this.get<Level[]>({ path: `/iep/levels/${agencyId}` })
-  }
-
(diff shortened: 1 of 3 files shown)


server/data/incentivesApi.ts
```diff
@@ -207,13 +207,6 @@ export class IncentivesApi extends RestClient {
     })
   }
 
-  /**
-   * @deprecated use getPrisonIncentiveLevels
-   */
-  getAvailableLevels(agencyId: string): Promise<Level[]> {
```
ushkarev, line 213: the `Leve` interface can be deleted too as it'll be unreferenced

xoen: Good point, removed here: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/626/commits/002d1c203a1cfadfb1b50d52b444fc5105d730cd

ashleygyngell (approved):
Nice 👍

PR #627, opened 2023-10-12
INC-1323: Upgrade to Node.js 20

xoen:
Node.js 20 is a few months old and it's becoming the next LTS (Long-term support) release in a couple of weeks (2023-10-24).

https://nodejs.dev/en/about/releases/

We don't need to upgrade it now, but if the upgrade is uneventful it may be a good idea to upgrade to it sooner rather than later.

Tests run fine.


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Stage: base image
-FROM node:18.18-bullseye-slim as base
+FROM node:20.8.0-bullseye-slim as base
```
ushkarev, line 2: `20.8-bullseye-slim` so that patch-level gets picked up automatically? like in the CI tags above

package.json
```diff
@@ -35,8 +35,8 @@
     "clean": "rm -rf assets/stylesheets build dist node_modules"
   },
   "engines": {
-    "node": "^18",
-    "npm": "^9"
+    "node": "^20",
+    "npm": "^10"
```
ushkarev, line 39: i’d recommend removing the `engine.npm` field in package.json cos i don’t think it’s useful and we don’t do anything special with npm

Dockerfile
xoen: oh I thought I used that one, I'll check if that tag exist and change if it's there 👍

ushkarev: i checked, it is!

package.json
xoen: Done here: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/627/commits/90c8f3cf2c4fc567d7f1516c109dfbddd3c07292

Dockerfile
xoen: Done here: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/627/commits/3c9e0dc0bb765180c7198f08e70cee79fbe167ac

ushkarev: you made it `20.8-bullseye` not `20.8-bullseye-slim` – is that deliberate?

package.json
ushkarev: i was suggesting getting rid of just `npm`, so that you'd still get a warning is using an unexpected `node` version. but frankly it's only a warning and doesn't _stop_ you using a version of node that we don't test on 🤷

xoen: This should be better then 🤞 : https://github.com/ministryofjustice/hmpps-incentives-ui/pull/627/commits/1add893923baf6238d6d5d3c5c2a918c7317b460

.husky/pre-commit
```diff
@@ -2,7 +2,7 @@
 . "$(dirname "$0")/_/husky.sh"
 
 export NODE_ENV=dev
-npx lint-staged && npx tsc && npm test
+npx lint-staged && npx tsc
```
ushkarev, line 5: 👀

ushkarev (approved):

.husky/pre-commit
xoen: I think run linting and tsc is a good compromise:
- they run fast
- they format code for you
- they can catch some bad stuff
- and most importantly, they don't fail when everything is ok

Tests are run by CI anyway. 

As mentioned before, I literally could not git rebase locally because of the tests failing ONLY when run by Husky, so I'd rather not have a tool be in my way while I'm trying to work - I can run tests manually and CI will run them anyway.

ushkarev: i want husky and lint-staged gone altogether :p

PR #646, opened 2023-11-02
INC-1324: Get user and roles from Manage Users API rather than from HMPPS Auth

xoen:
Cherry picked commit from TS template and tweaked/fixed integration tests.

### DIFF
--- a/feature.env
+++ b/feature.env
@@ -9,6 +9,7 @@ S3_SECRET_ACCESS_KEY=incentives
 DPS_URL=http://localhost:3000
 NOMIS_USER_ROLES_API_URL=http://localhost:9091/nomisUserRolesApi
 HMPPS_AUTH_URL=http://localhost:9091/auth
+MANAGE_USERS_API_URL=http://localhost:9091/manage-users-api
 TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
 TOKEN_VERIFICATION_ENABLED=true
 HMPPS_INCENTIVES_API_URL=http://localhost:9091/incentivesApi
--- a/integration_tests/mockApis/manageUsersApi.ts
+++ b/integration_tests/mockApis/manageUsersApi.ts
@@ -1,6 +1,7 @@
 import { Response } from 'superagent'
 
 import { stubFor } from './wiremock'
+import { UserRole } from '../../server/data/manageUsersApiClient'
 
 const stubUser = (name: string) =>
   stubFor({
@@ -22,7 +23,7 @@ const stubUser = (name: string) =>
     },
   })
 
-const stubUserRoles = () =>
+const stubUserRoles = (roles: UserRole[] = [{ roleCode: 'SOME_USER_ROLE' }]) =>
   stubFor({
     request: {
       method: 'GET',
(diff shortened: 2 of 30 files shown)


feature.env
```diff
@@ -9,6 +9,7 @@ S3_SECRET_ACCESS_KEY=incentives
 DPS_URL=http://localhost:3000
 NOMIS_USER_ROLES_API_URL=http://localhost:9091/nomisUserRolesApi
 HMPPS_AUTH_URL=http://localhost:9091/auth
+MANAGE_USERS_API_URL=http://localhost:9091/manage-users-api
```
ushkarev, line 12: this is the same as `NOMIS_USER_ROLES_API_URL`

xoen: I don't think so, the URL is definitely different for example: 

```
NOMIS_USER_ROLES_API_URL: "https://nomis-user-roles-api-dev.prison.service.justice.gov.uk"
MANAGE_USERS_API_URL: "https://manage-users-api-dev.hmpps.service.justice.gov.uk"
```

I confess these APIs and the way they're named confuses me (but maybe it's just because I'm not too involved into these).

We use the NomisUserRoles API to get the user case loads, while the Manage Users API is now used to get the user and user roles (instead of taking using 2 now deprecated HMPPS Auth API endpoints)

If it makes sense 🙃

ushkarev: but should the "manage users" one be used in place of "nomis user roles" one?

xoen: I've checked the swagger docs and the two APIs are not the same. Specifically, the NOMIS User Roles API endpoint we're using to get the user caseloads (`GET /me/caseloads`) is _**not there**_ in Manage Users API.

So short answer is no, it's not a drop-in replacement and we can't use Manage Users API to do the same thing.

xoen:
FYI: Rebased.

integration_tests/mockApis/manageUsersApi.ts
```diff
@@ -0,0 +1,43 @@
...
+      jsonBody: {
+        staffId: 231232,
+        username: 'USER1',
+        active: true,
+        name,
+      },
+    },
+  })
+
+const stubUserRoles = () =>
```
ushkarev, line 25: the old `stubUserRoles` used to allow passing in an optional list of roles, but this one does not.

it looks like there's now [a test](https://github.com/ministryofjustice/hmpps-incentives-ui/pull/646/files#diff-db80e94f69111242b7dbaf1b18f51d011296fa55b823d80c66e948af17d54deaR12-R15) that is using `stubSignIn` & `stubUserRoles` incorrectly: it passes override roles to `stubSignIn` (which doesn't accept them)

ushkarev: found [another test too](https://github.com/ministryofjustice/hmpps-incentives-ui/pull/646/files#diff-db80e94f69111242b7dbaf1b18f51d011296fa55b823d80c66e948af17d54deaR12-R17)

server/services/userService.ts
```diff
@@ -1,20 +1,20 @@
 import { convertToTitleCase } from '../utils/utils'
-import type HmppsAuthClient from '../data/hmppsAuthClient'
 import { NomisUserRolesApi, type Caseload } from '../data/nomisUserRolesApi'
+import ManageUsersApiClient from '../data/manageUsersApiClient'
 
 export interface UserDetails {
```
ushkarev, line 5: shouldn't this extend `User` from `'../data/manageUsersApiClient'` cos it's calling getUser and adding 3 new fields to it?

xoen: Good point, done it here: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/646/commits/381be600ac783231e4ed439eb8b881d357ecb8d0

integration_tests/mockApis/manageUsersApi.ts
xoen: Done here, hope this is what you had in mind: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/646/commits/17db7d789466101f5b6674ebb30a9a21e5986032

ushkarev (approved):

integration_tests/mockApis/manageUsersApi.ts
ushkarev: yup! i was wrong about `stubSignIn` (i was looking at a different method) but am happy with the new `stubUserRoles`!

feature.env
ushkarev: long answer and new plan _will_ be:
• the template will be updated to use the token to look up the current user’s roles (we will need to cherry-pick again!)
• we should (maybe) use the prison api to look up a user’s case loads (the NOMIS User Roles api is supposed to be more internal to feed other app-facing apis)

PR #649, opened 2023-11-06
Move `preprod` to use shared IP address allow lists

ushkarev:
…keeping `prod` unchanged for now to test the set actually applied to the ingress.

Expectation is that 5 SSCL ranges will disappear, but they are very likely to not be needed.

### DIFF
--- a/helm_deploy/values-preprod.yaml
+++ b/helm_deploy/values-preprod.yaml
@@ -10,6 +10,17 @@ generic-service:
   ingress:
     host: incentives-ui-preprod.hmpps.service.justice.gov.uk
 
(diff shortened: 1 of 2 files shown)


helm_deploy/values-preprod.yaml
```diff
@@ -10,6 +10,17 @@ generic-service:
   ingress:
     host: incentives-ui-preprod.hmpps.service.justice.gov.uk
 
+  allowlist:
+    groups:
+      - internal
+      - prisons
+      - private_prisons
```
xoen, line 17: I like this level of abstraction :)

PR #671, opened 2023-11-28
INC-1335: Display default image when user is not authorised to view a prisoner’s photo

ushkarev:
`prison-api` now (sensibly!) returns 403 instead of 404 if token is unauthorised. We should swallow the error and display a placeholder rather than a broken image; same as is done when none is found.

### DIFF
--- a/server/data/prisonApi.ts
+++ b/server/data/prisonApi.ts
@@ -25,11 +25,17 @@ export class PrisonApi extends RestClient {
     super('HMPPS Prison API', config.apis.hmppsPrisonApi, token)
   }
 
-  getImageByPrisonerNumber(prisonerNumber: string): Promise<Buffer> {
-    return this.get({
+  getImageByPrisonerNumber(prisonerNumber: string): Promise<Buffer | null> {
+    return this.get<Buffer>({
       path: `/api/bookings/offenderNo/${encodeURIComponent(prisonerNumber)}/image/data`,
       query: { fullSizeImage: 'false' },
-      handle404: true,
+    }).catch(error => {
+      const status = error?.status
+      if (status === 403 || status === 404) {
+        // return null if unauthorised or not found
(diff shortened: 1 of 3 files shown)


server/data/restClient.ts
```diff
@@ -74,9 +73,6 @@ export default class RestClient {
       return raw ? result : result.body
     } catch (error) {
       const sanitisedError = sanitiseError(error)
-      if (handle404 === true && error?.response?.status === 404) {
```
xoen, line 77: Was this only used by that method? May be useful for other methods? I guess we can remove if it's not used.

xoen (approved):

server/data/prisonApi.ts
```diff
@@ -25,11 +25,17 @@ export class PrisonApi extends RestClient {
...
       path: `/api/bookings/offenderNo/${encodeURIComponent(prisonerNumber)}/image/data`,
       query: { fullSizeImage: 'false' },
-      handle404: true,
+    }).catch(error => {
+      const status = error?.status
+      if (status === 403 || status === 404) {
+        // return null if unauthorised or not found
+        return null
+      }
+      throw error
```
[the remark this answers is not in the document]
ushkarev: i figured that is the photo is genuinely not found or "you" aren't allowed to see it, then a placeholder makes sense. if there's a random error, i thought it would be helpful to bubble it up just in case something bad is going on.

server/data/restClient.ts
ushkarev: it was only used in prison-api class and so not applicable anymore since we need to "handle" 403 and 404 together. i think trapping 404s _could_ indeed be useful in general. maybe we can have a go at putting something like that into the template’s `RestClient` – but maybe give it a more descriptive name, like `returnNullOn404` or something

PR #678, opened 2024-01-03
INC-1337: Copy incentive level history and recording pages from DPS

ashleygyngell:


server/utils/nunjucksSetup.ts
```diff
@@ -63,6 +63,54 @@ export default function nunjucksSetup(app: express.Express): void {
...
+    return null
+  })
+
+  njkEnv.addFilter('showDefault', (value, specifiedText) => {
+    if (value === 0) return value
+
+    return value || specifiedText || '--'
+  })
+
+  njkEnv.addFilter('addDefaultSelectedVale', (items, text, show) => {
```
xoen, line 84: is this supposed to be `addDefaultSelectedValue` (as in "value") or am I missing something?

assets/js/incentiveLevel.js
```diff
@@ -0,0 +1,14 @@
...
+  // record incentive level confirmation page
+  $('.govuk-button').on('click', function (e) {
+    const $prompt = $(e.target)
+    const gaId = $prompt.data('ga-id')
+    const caseload = $prompt.data('case-load')
+    if (gaId && typeof gtag === 'function') {
+      gtag('event', 'click', {
+        event_category: gaId,
+        event_label: caseload,
+      })
```
ushkarev, line 11: the particular details of the event sent to google analytics need to be adapted to make sense in the context of _this_ app. these likely made sense on dps, but not so much here cos different names are used.

so basically, the purpose is to track which of the 2 links are clicked on the confirmation page. what i'd do is adapt this `gtag` call to be more like what’s in `/assets/js/gaSendEvents.js#L20`

1) the active case load is already set globally so you can get rid of the `data-case-load` attribute in the template and the js code here that pulls it out
2) i would hazard a suggestion to rename the event from `click` to `incentives_event` to match the other events sent by this site. @xoen do you agree? using the same name would make selecting and filtering easier i presume.
3) instead of `event_category: gaId,` we should have 2 rows with [category and label](https://github.com/ministryofjustice/hmpps-incentives-ui/blob/main/assets/js/gaSendEvents.js#L21-L22) – where the `category` should be the _same_ for both buttons (something like "Record incentive level confirmation") and the `label` is what distinguishes them (e.g. "manage reviews" and "case notes")

server/views/pages/prisonerIncentiveLevelDetails.njk
```diff
@@ -0,0 +1,180 @@
...
+{% from "components/datePicker/datePicker.njk" import datePicker %}
+
+{% set title =  "Incentive details" %}
+
+{% block beforeContent %}
+  {{ govukBreadcrumbs({
+    items: [
+      {
+        text: "Home",
+        href: '/'
```
ushkarev, line 18: minor point – you can leave all of these as-is!
for the sake of consistency, we try to use only `"` in templates (to match how html attributes are wrapped) and `'` in typescript/javascript/stylesheets (cos that's the style dps seems to go for). the theory being that it's a bit easier to read if only one type is used throughout a file.

server/config.ts
```diff
@@ -111,6 +111,15 @@ export default {
       },
       agent: new AgentConfig(Number(get('HMPPS_PRISON_API_TIMEOUT_RESPONSE', 10000))),
     },
+    offenderSearchApi: {
+      url: get('OFFENDER_SEARCH_API_URL', 'http://localhost:8082', requiredInProduction),
```
ushkarev, line 115: port 8082 is already in use by another api so you can go for 8083… but in practice, none of us runs this site locally against services in docker so it doesn't really matter.

this fallback settings is _supposed_ to be for running the site along with everything else it needs in docker compose. we've never managed to do that cos of auth issues so we all use services in `dev` instead by actually setting the environment variables. so really, this fallback setting isn't necessary – tho for consistency, might as well have 8083 i guess

package.json
```diff
@@ -101,6 +101,7 @@
     "@aws-sdk/client-s3": "^3.460.0",
     "@ministryofjustice/frontend": "^1.8.1",
     "@sentry/node": "^7.82.0",
+    "@types/qs": "^6.9.11",
```
ushkarev, line 104: this ought to go into the dev dependencies section. anything to do with typescript compilation and validation should go there as it doesn't remain in the real running application

server/data/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Users without global search cannot search outside their active caseload nor link out to profiles outside their caseload set
+ */
+export const userRoleGlobalSearch = 'ROLE_GLOBAL_SEARCH' as const
+
+/**
+ * Users without inactive bookings cannot search for released prisoners nor link out to profiles of those individuals
+ */
+export const userRoleInactiveBookings = 'ROLE_INACTIVE_BOOKINGS' as const
```
ushkarev, line 20: these 3 constants aren't needed in this pr cos these roles don't affect the access permissions.

what you could do (if you wanted) is add `ROLE_MAINTAIN_IEP` here and use it as a constant in place of that string elsewhere.

server/data/incentivesApi.ts
```diff
@@ -122,11 +122,52 @@ export interface IncentivesReview {
...
     super('HMPPS Incentives API', config.apis.hmppsIncentivesApi, systemToken)
   }
 
+  getIncentiveSummaryForPrisoner(prisonerNumber: string): Promise<IncentiveSummaryForBookingWithDetails> {
+    return this.get({ path: `/incentive-reviews/prisoner/${encodeURIComponent(prisonerNumber)}` })
+  }
+
+  updateIncentiveLevelForPrisoner(prisonerNumber: string, data: IepLevelChangeRequest, ): Promise<IepLevelChangeRequest> {
+    return this.post({
+      path: `/incentive-reviews/prisoner/${prisonerNumber}`,
```
ushkarev, line 166: to be safer, you should put `prisonerNumber` inside encodeURIComponent as in the method above

server/routes/prisonerChangeIncentiveLevelDetails.ts
```diff
@@ -0,0 +1,142 @@
...
+      // })
+      // raiseAnalyticsEvent(
+      //   'Update Incentive level',
+      //   `Level changed from ${iepLevel} to ${newIepLevel} at ${agencyId}`,
+      //   'Incentive level change'
+      // )
+      return renderConfirmation(req, res)
+    } catch (error) {
+      // TODO: CHECK URL
+      res.locals.redirectUrl = `/prisoner/${offenderNo}`
```
ushkarev, line 133: this will need to redirect to the new prisoner profile on DPS so we would need a whole new environment variable like `PRISONER_PROFILE_URL: "https://prisoner-dev.digital.prison.service.justice.gov.uk"`

server/routes/all.ts
```diff
@@ -24,6 +26,16 @@ export default function routes(userService: UserService): Router {
     authorisationMiddleware([manageIncentiveLevelsRole]),
     incentiveLevelRoutes(standardRouter(userService)),
   )
+  router.use(
+    '/incentive-reviews/prisoner',
+    authorisationMiddleware(['ROLE_MAINTAIN_IEP']),
```
ushkarev, line 31: does a user require the `ROLE_MAINTAIN_IEP` role just to see incentive reviews? i think it's only needed to update them so this and line 36 should go.

server/routes/prisonerChangeIncentiveLevelDetails.ts
ushkarev: i just noticed you have https://github.com/ministryofjustice/hmpps-incentives-ui/pull/678/files#diff-7c6cd32da5f152ddbb8d8312c23cd3b5f1e15ee37e3d706ca81344ff6003eebdR210 which is similar and uses the old url (which works cos it's redirected) but maybe both ought to use the new urls…?

server/utils/raiseAnalyticsEvent.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (!config.googleAnalytics.ga4MeasurementId) return Promise.resolve()
+  const ga = ua(config.googleAnalytics.ga4MeasurementId)
+  const data = {
+    ec: category,
+    ea: action,
+    el: label,
+    ev: value,
+  }
+  return ga.event(data).send()
+}
```
ushkarev, line 15: i think this whole thing should be removed (and references to it) because we do not do this for any other server-side actions and it won't work anyway as it's trying to use google universal analytics rather than google analytics 4

server/config.ts
ashleygyngell: That would make sense why it wasn't flagging up as already used. Thank you for catching that - missed that port on frontendComponents.

server/data/constants.ts
ashleygyngell: Thanks - had brought these over but wasn't sure if i would need them down the line. Agree to adding `ROLE_MAINTAIN_IEP` here.

server/routes/all.ts
ushkarev: error, line 36 should remain

server/data/prisonApi.ts
```diff
@@ -60,15 +77,20 @@ export class PrisonApi extends RestClient {
...
-    return this.get<Agency>({
+  getStaffDetails(context: string, staffId: string): Promise<Staff> {
+    return this.get<Staff>({
       path: `/api/users/${staffId}`,
-      query: context
+      query: context,
     })
   }
 
+  getDetails(prisonerNumber: string): Promise<Offender> {
```
xoen, line 99: You may want to rename this to `getPrisonerDetails()` or something like that, for clarity and also for consistency with `getStuffDetails()`.

server/routes/prisonerIncentiveLevelDetails.ts
```diff
@@ -65,109 +63,162 @@ const SYSTEM_USERS = ['INCENTIVES_API']
...
-            }
-            return prisonApi.getStaffDetails(systemToken, userId)
-          })
-      )
-    ).filter(notEmpty)
 
-    const establishments: any[] = await Promise.all(uniqueAgencyIds.filter((id) => Boolean(id)).map((id) => prisonApi.getAgencyDetails(systemToken, id))
-    )
-    const levels = Array.from(new Set(incentiveLevelDetails.iepDetails.map((details) => details.iepLevel))).sort()
+    try {
```
xoen, line 75: This is a bit try block, what's the rational for this? Just redirect on any error? Is this what DPS was doing?

assets/js/incentiveLevel.js
xoen: yes, you could track this by binding the on click event like done in here: https://github.com/ministryofjustice/hmpps-incentives-ui/blob/9f8be5b8ff32678665b4ecbccbc612e34c80200b/assets/js/gaSendEvents.js#L6

server/routes/prisonerChangeIncentiveLevelDetails.ts
```diff
@@ -9,94 +9,96 @@ import {RequestHandler, Router} from "express";
...
+                  ? `/incentive-summary/${agencyId}-${locationId.split('-')[0]}`
+                  : '/',
           nextReviewDate: nextReviewDate?.format('D MMMM YYYY'),
+          prisonerNumber,
+          prisonerName: formatName(firstName, lastName),
+          profileUrl: `${res.app.locals.dpsUrl}/prisoner/${prisonerNumber}`,
         })
       } catch (error) {
-        res.locals.redirectUrl = `/prisoner/${prisonerNumber}/incentive-level-details`
+        res.locals.redirectUrl = `incentive-level-details/prisoner/${prisonerNumber}`
```
xoen: does this also need a leading `/` as well?

assets/js/incentiveLevel.js
```diff
@@ -1,13 +1,15 @@
...
-        event_category: gaId,
-        event_label: caseload,
+    const gaCategory = $prompt.data('ga-category')
+    const gaAction = $prompt.data('ga-action')
+
+    if (gaCategory && typeof gtag === 'function') {
+      gtag('event', 'incentives_event', {
+        category: gaCategory,
+        action: gaAction,
       })
```
xoen: I think `gaSendEvent` would do pretty much the same thing, but I haven't touched this in awhile so double-check: https://github.com/ministryofjustice/hmpps-incentives-ui/blob/9f8be5b8ff32678665b4ecbccbc612e34c80200b/assets/js/gaSendEvents.js

xoen: you may be able to simplify with something along the lines of `$('.govuk-button').on('click', gaSendEvent)`

ushkarev: `gaSendEvent` is not loaded on this page. It _could_ be but it has separate initialisation which would need a tweak. I thought at this point the two could be separate – we don't have any nice front end JS bundling in our projects

package.json
```diff
@@ -119,17 +119,20 @@
...
     "jwt-decode": "^4.0.0",
+    "moment": "^2.30.1",
     "nocache": "^4.0.0",
     "nunjucks": "^3.2.4",
     "passport": "^0.7.0",
     "passport-oauth2": "^1.7.0",
     "prom-client": "^15.1.0",
     "redis": "^4.6.12",
     "superagent": "^8.1.2",
+    "universal-analytics": "^0.5.3",
```
ushkarev, line 130: we can remove this dependency since we're not gonna raise analytics events server-side

(ps: google analytics 4 has replaced google universal analytics)

server/routes/prisonerIncentiveLevelDetails.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    })
+
+    return request(app)
+      .get(`/incentive-reviews/prisoner/${prisonerNumber}`)
+      .set('authorization', `bearer ${tokenWithNecessaryRole}`)
+      .expect(200)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Record incentive level')
+        expect(res.text).toContain('Record incentive level')
```
ushkarev, line 174: duplicate line

server/routes/prisonerIncentiveLevelDetails.test.ts
```diff
@@ -0,0 +1,295 @@
...
+      .set('authorization', `bearer ${tokenWithMissingRole}`)
+      .expect(200)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Incentive level history')
+        expect(res.text).not.toContain('Record incentive level')
+      })
+  })
+
+  // TODO: Check setup of caseloads and role. Dont think set authorization is used here
```
ushkarev, line 190: is this todo still relevant?

server/routes/prisonerChangeIncentiveLevelDetails.test.ts
```diff
@@ -0,0 +1,324 @@
...
+        active: true,
+      })
+    }
+    return Promise.resolve({
+      agencyId: 'LEI',
+      description: '123',
+      agencyType: '123',
+      active: true,
+    })
+  })
```
ushkarev, line 134: from what i can tell, `prisonApi.getStaffDetails` and `prisonApi.getAgency` are not called by this page

server/routes/prisonerChangeIncentiveLevelDetails.test.ts
```diff
@@ -0,0 +1,324 @@
...
+    const error: SanitisedError = {
+      name: 'Error',
+      status: 404,
+      message: 'Not Found',
+      stack: 'Not Found',
+    }
+    offenderSearch.getPrisoner.mockRejectedValue(error)
+    return request(app)
+      .get(`/incentive-reviews/prisoner/${prisonerNumber}`)
+      .expect(404)
```
ushkarev, line 181: we should still include the valid authorisation bearer token so that it is purely the missing prisoner that causes this request to 404

server/routes/prisonerChangeIncentiveLevelDetails.test.ts
```diff
@@ -0,0 +1,324 @@
...
+          reason: 'Should retain this text',
+        })
+        .expect(200)
+        .expect(res => {
+          expect(res.text).toContain('Select an incentive level, even if it is the same as before')
+          expect(res.text).toContain('Should retain this text')
+        })
+    })
+
+    it('should return generic error page if api returns an error', () => {
```
ushkarev, line 272: good idea to test this, actually. i would recommend something other than a 404 though – there are too many other potential causes of a 404. maybe a 400 just to be different. i would have it only on the `updateIncentiveLevelForPrisoner` so that we're pretending the actual form being posted was the cause of the error.

package.json
```diff
@@ -119,17 +119,20 @@
...
     "passport-oauth2": "^1.8.0",
     "prom-client": "^15.1.0",
     "redis": "^4.6.12",
     "superagent": "^8.1.2",
+    "universal-analytics": "^0.5.3",
     "url-value-parser": "^2.2.0",
     "uuid": "^9.0.1"
   },
   "devDependencies": {
+    "@govuk-react/date-field": "^0.10.6",
```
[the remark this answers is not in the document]
ashleygyngell: Nothing ha! Leftover from when i was playing around with getting the new date picker to work.

assets/scss/components/_mini-profile-header.scss
```diff
@@ -0,0 +1,50 @@
...
+      &:last-child {
+        margin-bottom: 0;
+
+        dd {
+          margin-bottom: 0;
+        }
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
ashleygyngell: Nope! No prisoner photo in this :)

server/routes/prisonerChangeIncentiveLevelDetails.ts
```diff
@@ -0,0 +1,144 @@
...
+      return res.render('pages/prisonerChangeIncentiveLevelConfirmation.njk', {
+        agencyId,
+        bookingId,
+        breadcrumbPrisonerName: putLastNameFirst(firstName, lastName),
+        cancelUrl: '/',
+        incentiveSummary,
+        manageIncentivesUrl:
+          agencyId && locationId && locationId.includes('-')
+            ? `/incentive-summary/${agencyId}-${locationId.split('-')[0]}`
+            : '/',
```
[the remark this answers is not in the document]
ashleygyngell: Yes good point` / ` just takes you back to incentives homepage which isn't where you would have started the journey. Not sure about `/select-location`. I mean its only a fall back but would it make more sense to return to the prisoners profile page?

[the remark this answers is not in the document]
ushkarev: but this link says “Return to incentive reviews”, it's the _other_ one that says “Return to prisoner profile page”! this one should get to as near as it can to the reviews table – in this case it cannot get to it directly so /select-location is the next best thing in my opinion

server/routes/prisonerChangeIncentiveLevelDetails.ts
```diff
@@ -0,0 +1,144 @@
...
+      const locationId: string | undefined = assignedLivingUnit?.description
+      const incentiveSummary = await incentivesApi.getIncentiveSummaryForPrisoner(prisonerNumber)
+      const nextReviewDate =
+        incentiveSummary?.nextReviewDate && moment(incentiveSummary.nextReviewDate, 'YYYY-MM-DD HH:mm')
+
+      return res.render('pages/prisonerChangeIncentiveLevelConfirmation.njk', {
+        agencyId,
+        bookingId,
+        breadcrumbPrisonerName: putLastNameFirst(firstName, lastName),
+        cancelUrl: '/',
```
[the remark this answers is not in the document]
ashleygyngell: I think i had set it up with the intention of using in the template. But think thats unnecessary.

server/views/pages/prisonerChangeIncentiveLevelDetails.njk
```diff
@@ -0,0 +1,98 @@
...
+      text: "Save",
+      type: "submit",
+      classes: "govuk-!-margin-right-2",
+      attributes: { "data-test": "submit-change" }
+    }) }}
+
+    {{ govukButton({
+      text: "Cancel",
+      element: 'a',
+      href: "/" ,
```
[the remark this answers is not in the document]
ashleygyngell: 100%

server/routes/prisonerChangeIncentiveLevelDetails.ts
```diff
@@ -0,0 +1,144 @@
...
+        currentIncentiveLevel,
+        errors,
+        formValues,
+        prisonerNumber,
+        prisonerName: formatName(firstName, lastName),
+        profileUrl: `${res.app.locals.dpsUrl}/prisoner/${prisonerNumber}`,
+        selectableLevels,
+      })
+    } catch (error) {
+      res.locals.redirectUrl = `incentive-level-details/prisoner/${prisonerNumber}`
```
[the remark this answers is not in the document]
ashleygyngell: ^ should be `incentive-reviews/prisoner/${prisonerNumber}` anyway.

server/routes/prisonerChangeIncentiveLevelDetails.ts
```diff
@@ -0,0 +1,144 @@
...
+  // @ts-expect-error
+  const index = async (req, res) => renderTemplate(req, res, undefined)
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  const postForm = async (req, res) => {
+    const { prisonerNumber } = req.params
+    const systemToken = await hmppsAuthClient.getSystemClientToken(res.locals.user.username)
+    const incentivesApi = new IncentivesApi(systemToken)
+    const { offenderNo } = req.params
```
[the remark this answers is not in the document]
ashleygyngell: Yes, thanks. I hate that prisonerNumber and offenderNo both exist as terms

server/routes/prisonerChangeIncentiveLevelDetails.test.ts
ashleygyngell: You would be correct!

package.json
ushkarev: we don't need this dependency

server/utils/utils.ts
```diff
@@ -9,11 +11,31 @@ const isBlank = (str: string): boolean => !str || /^\s*$/.test(str)
...
+  if (firstName && !lastName) return properCaseName(firstName)
+
+  return `${properCaseName(lastName)}, ${properCaseName(firstName)}`
+}
+
+export const possessive = (string: string): string => {
+  if (!string) return ''
+
+  return `${string}${string.toLowerCase().endsWith('s') ? '’' : '’s'}`
+}
```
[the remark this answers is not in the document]
ushkarev: did you want a hand with some of the unit tests for the new utils?

server/routes/prisonerChangeIncentiveLevelDetails.test.ts
```diff
@@ -0,0 +1,298 @@
...
+      message: 'Not Found',
+      stack: 'Not Found',
+    }
+    offenderSearch.getPrisoner.mockRejectedValue(error)
+    return request(app)
+      .get(`/incentive-reviews/prisoner/${prisonerNumber}/change-incentive-level`)
+      .set('authorization', `bearer ${tokenWithNecessaryRole}`)
+      .expect(200)
+  })
+})
```
ushkarev, line 158: i think this test isn't quite right. from what i can tell, the "record incentive level" page doesn't call offender search so really
1) offender search mocking ought to be removed from this file
2) the 404 could instead be done on `prisonApi.getPrisonerDetails` and then i think you can once again have something like this, as you did in the past
```
      .expect(404)
      .expect(res => {
        expect(res.text).toContain('Page not found')
        expect(res.text).not.toContain('John, Smith')
      })
```

ushkarev:
oh and it looks like some of the tests actually fail :(

```
 FAIL  server/routes/prisonerChangeIncentiveLevelDetails.test.ts
  ● POST /incentive-reviews/prisoner/change-incentive-level › When there are errors › should return generic error page if api returns an error

 FAIL  server/routes/prisonerIncentiveLevelDetails.test.ts
  ● GET /incentive-reviews/prisoner/ › should return 404 if prisoner is not found
```

ashleygyngell:
@xoen Thanks for giving this a look - in hindsight, setting up PR's for each page would have been best for clarity, agreed!

xoen:
@ashleygyngell yeah probably but don't worry, it's a big task regardless of how you split it :)

PR #683, opened 2024-01-09
INC-1335: Improve integration tests

ushkarev:
Previously, stubbing did not work for DPS’ frontend components api as the url was never set to point to wiremock.

Now, in most tests the fallback header and footer is used (to make signing in possible) but a new test exists to ensure html, css and javascript is correctly loaded.


feature.env
```diff
@@ -14,6 +16,7 @@ TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
 TOKEN_VERIFICATION_ENABLED=true
 HMPPS_INCENTIVES_API_URL=http://localhost:9091/incentivesApi
 HMPPS_PRISON_API_URL=http://localhost:9091/prisonApi
+COMPONENT_API_URL=http://localhost:9091/frontendComponents
```
ushkarev, line 19: ☝️ without this, frontend component stubbing had no effect

PR #714, opened 2024-02-28
INC-1352: Use stricter type-checking in integration tests

ushkarev:
…which leads to needing more-realistic samples and type corrections


ashleygyngell (approved):
👍

PR #721, opened 2024-03-06
INC-1354: Drop date picker from incentive level history page

ushkarev:
…because it is not particularly accessible and messes up form submission when using a keyboard.

This commit can be reverted in future as a starting point when the component is improved across DPS / MoJ sites.


ashleygyngell (approved):

ushkarev:
Not needed; going with date picker as-is

PR #803, opened 2024-07-15
INC-1325: Remove HotJar tracking

ushkarev:
User actions would now _not_ be recorded on the charts (/analytics) and info pages


xoen (approved):
👍👍👍

PR #823, opened 2024-08-16
INC-1375: Swap old date-picker copied from DPS for standard MoJ component

ushkarev:
Ref: https://design-patterns.service.justice.gov.uk/components/date-picker/


assets/scss/components/_horizontal-content.scss
```diff
@@ -44,7 +44,7 @@
     }
 
     > .govuk-button-group {
-      align-self: flex-end;
+      align-self: center;
```
xoen, line 43: 🌟

xoen (approved):

assets/scss/components/_horizontal-content.scss
ushkarev: god knows why this worked

PR #834, opened 2024-08-30
INC-1335: Send alerts to dev channel

xoen:
Send them to the `#incentives-dev` Slack channel.

### DIFF
--- a/helm_deploy/hmpps-incentives-ui/values.yaml
+++ b/helm_deploy/hmpps-incentives-ui/values.yaml
@@ -63,4 +63,3 @@ generic-service:
 
 generic-prometheus-alerts:
   targetApplication: hmpps-incentives-ui
-  alertSeverity: digital-prison-service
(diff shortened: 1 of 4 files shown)


helm_deploy/hmpps-incentives-ui/values.yaml
```diff
@@ -63,4 +63,3 @@ generic-service:
 
 generic-prometheus-alerts:
   targetApplication: hmpps-incentives-ui
-  alertSeverity: digital-prison-service
```
xoen, line 66: The default value doesn't matter but it was pointing to `#dps_alerts`.

I've removed it: [The default in the `generic-prometheus-alerts` helm chart](https://github.com/ministryofjustice/hmpps-helm-charts/blob/ab502d62c7b371ee8930acc1abdd80b1feec2065/charts/generic-prometheus-alerts/values.yaml#L13) is to send them to the `#dps_alerts_non_prod` which is a more sensible default.

ushkarev (approved):

PR #883, opened 2025-02-03
Fixes VB-4369 - improves guidance on VOs linked to incentive levels

simonnebMOJ:
The number of visiting orders (and privileged visiting orders) a prisoner receives for a given incentive level at a prison is now controlled in the incentives API.

The information presented to staff as they input those numbers is very limited - this means that some prisons have inputted very high numbers in that service. We also get queries from staff asking what the allowances are and when they will get added.

This PR improves the information shown to staff, including:
* how frequently VOs get updated
* how frequently PVOs get updated
* if and when the VOs or PVOs ‘expire'

By improving the information in the Incentive service, we can help staff understand what the VO and PVO numbers will mean for their establishment and prisoners in that establishment.

### DIFF
--- a/server/views/partials/prisonIncentiveLevelEditPartial.njk
+++ b/server/views/partials/prisonIncentiveLevelEditPartial.njk
@@ -117,7 +117,7 @@
     id: form.formId + "-" + fieldId,
     name: fieldId,
     value: field.value,
-    suffix: {text: "added every 14 days"},
+    suffix: { text: "added every 14 days" },
     inputmode: "numeric",
     spellcheck: false,
     errorMessage: { text: field.error } if field.error else undefined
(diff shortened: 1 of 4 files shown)


server/views/partials/prisonIncentiveLevelEditPartial.njk
```diff
@@ -117,6 +117,7 @@
     id: form.formId + "-" + fieldId,
     name: fieldId,
     value: field.value,
+    suffix: { text: "added every 14 days" },
```
ushkarev, line 120: did you not want the "per week" suffix for the 2 transfer limits higher up to match the view page?

ushkarev (dismissed):

PR #897, opened 2025-03-07
Secure access to incentive reviews page

ushkarev:
…by checking user caseloads and roles with respect to the prisoner’s current location:

• if person is in transfer, global search role is required
• if person has been released, inactive bookings role is required
• otherwise person’s current establishment must be in user’s caseloads


Mjwillis (approved):
LGTM

PR #917, opened 2025-05-12
IR-1027: Backported TypeScript template changes

xoen:
Mainly the use of classes from the `@ministryofjustice/hmpps-rest-client` library.


package.json
```diff
@@ -85,6 +82,9 @@
     "@aws-sdk/client-s3": "^3.806.0",
     "@aws-sdk/client-sqs": "^3.806.0",
     "@ministryofjustice/frontend": "^5.1.3",
+    "@ministryofjustice/hmpps-auth-clients": "^0.0.1-alpha.4",
+    "@ministryofjustice/hmpps-monitoring": "^0.0.1-beta.2",
+    "@ministryofjustice/hmpps-rest-client": "^0.0.1-alpha.9",
     "agentkeepalive": "^4.6.0",
```
ushkarev, line 88: don't need `agentkeepalive` and `superagent` anymore, right?

xoen: I thought I saw some imports for both, but I think it's just `agentkeepalive` that can go...lemme switch context and have a look in a sec.

xoen: FYI: I've removed the unused dependency to `agentkeepalive` here: https://github.com/ministryofjustice/hmpps-incentives-ui/pull/917/commits/503f8ce228b4417e711a9e60b110289e6eea921e

But we import `superagent`'s stuff in a few places, [e.g. here](https://github.com/ministryofjustice/hmpps-incentives-ui/blob/2978e56a1c81a1e7921705b77f62b9f446267591/integration_tests/mockApis/auth.ts#L1). They would still work as `superagent` is a transient dependency anyway, but I think it's better to be explicit and depend on it directly given we use it in our codebase. IMHO anyway.

ushkarev (approved):

PR #920, opened 2025-05-14
Set GIT_BRANCH environment variable

xoen:
This is needed by the `/info` endpoint and to be able to deploy.


ushkarev (approved):
if i were you, i'd copy the whole Dockerfile from the other projects to make it easier to copy future changes

PR #934, opened 2025-06-20
Migrate to GH Actions (from CircleCI)

xoen:


ushkarev (approved):

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,183 @@
...
+    uses: ministryofjustice/hmpps-github-actions/.github/workflows/node_unit_tests.yml@v2 # WORKFLOW_VERSION
+    needs: [node_build]
+    secrets: inherit
+  # generic node integration tests using wiremock - feel free to override with local tests if required
+  node_integration_tests:
+    runs-on: ubuntu-latest
+    needs: [node_build]
+    services:
+      redis:
+        image: redis:7.0
```
ushkarev, line 53: if redis can be used as a service in external workflows, why can't minio?

xoen: I tried but it's another limitation of GH Actions where you cannot override the command as we need in the case of MinIO.

ushkarev: it’s pretty annoying that GHA doesn’t let you change the container command or even pass arguments :/
`options` is actually for the `docker create …` command!

xoen: Yep.

PR #1117, opened 2026-05-27
Fix `qs` vulnerability (`moderate`)

xoen:
**NOTE**: There is a new version of Cypress that may depend on the non-vulnerable version of `qs` - but it was just released. Once we upgrade to the newer version of `cypress` the entry under `overrides` may no longer be needed and we can remove it.

See: https://github.com/advisories/GHSA-q8mj-m7cp-5q26

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2712,19 +2712,6 @@
         "node": "22 || 24 || 26"
       }
     },
-    "node_modules/@ministryofjustice/hmpps-auth-clients/node_modules/@ministryofjustice/hmpps-rest-client": {
-      "version": "1.2.1",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -2712,19 +2712,6 @@
         "node": "22 || 24 || 26"
       }
     },
-    "node_modules/@ministryofjustice/hmpps-auth-clients/node_modules/@ministryofjustice/hmpps-rest-client": {
```
xoen, line 2715: I think this was just vestigial and not used anywhere, npm just decided to catch-up and remove the older version from `package-lock.json`.

jonesj2193 (approved):

